PR #233, opened 2014-10-03
Colorization and alignment improvements from DragonFly branch

jrmarino:
As discussed on IRC (see title)


bdrewery:
Several pieces of this have been or are being implemented.  Other pieces (like removing the log output) will not be implemented. Thanks!

PR #248, opened 2014-12-03
Allow hooks to color

davehayes:
All this does is allow you to change the colors via a hook if you don't like them.


bdrewery:
Your change was not needed. I reworked the inclusion of colors.sh which now allows overriding any of the variables in poudriere.conf.

PR #252, opened 2014-12-09
Fixed passing arguments in shell script in (unlikely) case of spaces in args

yurivict:


bdrewery:
Merged 44f26ebf37aa574de856473c7a67b7b595569b57

PR #253, opened 2014-12-09
Switch MIPS targets to new ELFTOOLCHAIN

seanbruno:
Instead of using WITH_DEBUG, use the new elf toolchain to strip and such.


so14k:
Merged, thanks!

PR #258, opened 2014-12-15
RFC: Add git support for jail command

splbio:
I have a patch here that adds preliminary support for git as a source for FreeBSD source tree.

**This is not intended to be taken as-is, as this is only for comment/suggestions.**

If there is interest and feedback I can rebase and fix this against master branch.

I can also see what else needs to be done, as I've only just barely started work on this.

(One trick might be to set --depth for the git clone to make it faster, I will look into that.)

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -445,7 +445,7 @@ install_from_git() {
 	fi
 	build_and_install_world
 
-	git_sha=$(${GIT_CMD} git rev-parse --short HEAD)
+	git_sha=$(${GIT_CMD} rev-parse --short HEAD)
 	setvar "${var_version_extra}" "${git_sha}"
 }
 
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -182,6 +182,9 @@ rename_jail() {
 update_jail() {
 	SRC_BASE="${JAILMNT}/usr/src"
 	METHOD=$(jget ${JAILNAME} method)
+	SCM_BRANCH=$(jget ${JAILNAME} scm_branch 2>/dev/null)
+	SCM_URL=$(jget ${JAILNAME} scm_url 2>/dev/null)
+	_jget ARCH ${JAILNAME} arch
 	if [ -z "${METHOD}" -o "${METHOD}" = "-" ]; then
 		METHOD="ftp"
 		jset ${JAILNAME} method ${METHOD}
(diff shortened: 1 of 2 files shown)


src/share/poudriere/jail.sh
```diff
@@ -226,6 +229,15 @@ update_jail() {
...
+	git*)
+		install_from_git version_extra
+		RELEASE=$(update_version "${version_extra}")
+		jset ${JAILNAME} scm_branch "${SCM_BRANCH}"
+		jset ${JAILNAME} scm_url "${SCM_URL}"
+		update_version_env "${RELEASE}"
+		make -C ${SRC_BASE} delete-old delete-old-libs DESTDIR=${JAILMNT} BATCH_DELETE_OLD_FILES=yes
+		markfs clean ${JAILMNT}
+		;;
 	svn*)
```
bdrewery: All of the special handling in the SVN branch will be need to be merged with the git changes as well.

src/share/poudriere/jail.sh
```diff
@@ -399,6 +411,44 @@ install_from_src() {
...
+	git+ssh) proto="git+ssh" ;;
+	git+file) proto="file" ;;
+	git) proto="git" ;;
+	esac
+	if [ ${UPDATE} -eq 0 ]; then
+		msg_n "Checking out the sources using git..."
+		${GIT_CMD} clone --depth 1 ${SCM_BRANCH:+-b ${SCM_BRANCH}}  ${proto}://${SCM_URL} ${SRC_BASE} || err 1 " fail"
+		echo " done"
+		if [ -n "${SRCPATCHFILE}" ]; then
+			msg_n "Patching the sources with ${SRCPATCHFILE}"
```
bdrewery: It's probably better to put patching into 1 function.

src/share/poudriere/jail.sh
bdrewery: Ignore this, the expanded code was hidden. I was looking at the wrong code.

bdrewery:
Seems generally fine. Can you also update poudriere.8?

Did you have a concern that should block this from going in?

src/share/poudriere/jail.sh
```diff
@@ -399,6 +411,44 @@ install_from_src() {
...
+	case ${METHOD} in
+	git+http) proto="http" ;;
+	git+https) proto="https" ;;
+	git+ssh) proto="git+ssh" ;;
+	git+file) proto="file" ;;
+	git) proto="git" ;;
+	esac
+	if [ ${UPDATE} -eq 0 ]; then
+		msg_n "Checking out the sources using git..."
+		${GIT_CMD} clone --depth 1 ${SCM_BRANCH:+-b ${SCM_BRANCH}}  ${proto}://${SCM_URL} ${SRC_BASE} || err 1 " fail"
```
rbgarga: Maybe --single-branch would be useful here to reduce cloned size

emaste:
@splbio will you rebase and update the pull request?

emaste:
Superseded by a21b04dec31649d3181cd3d43cf8ae54ad5194a4 it seems

PR #261, opened 2014-12-23
Fix percent display

AMDmi3:
Since floating point value representing percent of finished ports is printed directly, it's usually some large string like 12.12512341% and is wrapped onto the next line of ```<span id="progress">```. Trim it to 2 digits after decimal point. While here, remove unused pctdonetxt variable.

### DIFF
--- a/src/share/poudriere/html/assets/poudriere.js
+++ b/src/share/poudriere/html/assets/poudriere.js
@@ -120,7 +120,7 @@ function determine_canvas_width() {
 
 function update_canvas(stats) {
 	var queued, built, failed, skipped, ignored, remaining, pctdone;
-	var height, width, x, context, canvas, pctdonetxt;
+	var height, width, x, context, canvas;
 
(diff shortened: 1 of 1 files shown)


bdrewery:
No, it's not printed directly. `Math.floor` is used to truncate off the decimal.

AMDmi3:
It is not. Floor output is assigned to pctdonetxt which is not used.

bdrewery:
Aha you're right.

bdrewery:
Your change makes it use 2 decimal places when none are intended. I will just fix it to use the proper `pctdonetxt`

AMDmi3:
Could 2 decimal places be added though? Usually lots more than 100 ports are built, and fractions of percent are useful for tracking progress.

bdrewery:
Maybe, it throws off all of the widths though. It's all very specifically sized. I hate this progress bar. It is painful to maintain.

PR #262, opened 2014-12-25
Fix progressbar painting

AMDmi3:
Widths of progressbar chunks are rounded down to closest percent for no apparent reason, so progress bard displays incorrect data - it won't show e.g. failed ports of there's <1% of them, it won't be fully filled when the build finishes. Simplify logic by removing useless rounding and thus fix progressbar.

### DIFF
--- a/src/share/poudriere/html/assets/poudriere.js
+++ b/src/share/poudriere/html/assets/poudriere.js
@@ -79,25 +79,15 @@ function format_pkgname(pkgname) {
 }
 
 function minidraw(x, height, width, context, color, queued, variable) {
-	var pct, total_pct, newx;
+	var newx;
 
-	/* Calculate how much percentage this value should display */
-	pct = Math.floor(variable * 100 / queued);
-	if (pct == 0) {
+	if (variable == 0) {
 		return 0;
 	}
-	newx = width * (pct / 100);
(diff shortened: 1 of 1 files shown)


bdrewery:
Note the comment, it's rounded down so it doesn't show 100% when building 24000 ports and you have 100 remaining.

AMDmi3:
There's no point in showing this with the bar. Percent count may be made to behave like that, e.g. ```Math.floor(percent * 100) / 100.0```

bdrewery:
This isn't quite right. It is not following the intent that 99% still shows a remaining bar. I agree the current behavior has the bugs you mention.

AMDmi3:
Is there really sense in this? I doubt anyone uses progress bar to determine if the build is finished - there's percent displayed just near it and there are many other indicator around. Even if the progress bar just adds one or several pixels gap, it'd be nearly visible with present resolutions.

bdrewery:
The sense is madness. I obsessively tried getting this progressbar perfect and it has had countless bugs. If the bar shows nothing remaining at 99% then another OCDish person like me will complain. We can't please everyone. I still think it makes general sense to round down before painting. I can live without the forced 99% special case. Doing that breaks the 0.5% case too though.

PR #270, opened 2015-01-26
Logging debug, verbose, and warning output.

heckendorfc:
Package dependencies and other information are displayed only through debug level messages. This information would be useful to have without cluttering the terminal output. This patch creates a log file for additional debug, verbose, and warning output during bulk runs.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -134,6 +134,7 @@ msg_n() {
 msg() { msg_n "$@"; echo; }
 
 msg_verbose() {
+	[ -n "${DEBUG_LOG}" ] && msg "$1" >> "${DEBUG_LOG}"
 	[ ${VERBOSE} -gt 0 ] || return 0
 	msg "$1"
 }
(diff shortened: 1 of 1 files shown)


src/share/poudriere/common.sh
```diff
@@ -3844,6 +3849,8 @@ prepare_ports() {
 		# Link this build as the /latest
 		ln -sfh ${BUILDNAME} ${log%/*}/latest
 
+		DEBUG_LOG=${log}/logs/debug.log
```
bdrewery: I think this should not be defined by default or only if WITH_DEBUG_LOGGING set or something. The msg_debug can be quite verbose and may impact performance by writing to this file.

I need to run this locally for a while to get a better idea on it though.

bdrewery:
By the way there is a bigger problem here in that the entire bulk console is not logged anywhere. The plan has been that normal usage of bulk would shift over to poudriered. That effort has stagnated a bit though. I do agree we need an easier way to get the debug log and the console log. Having the console log available from the web interface has been requested a few times.

PR #276, opened 2015-02-26
Move .cpignore handling to just the TMPFS_ALL case

QueuingKoala:
This avoids manipulating .cpignore files in cases where cpdup(1) is not
used. Only this case in clonefs() results in the requirement to have
cpdup-supporting files, and it can be assumed such writes will succeed
when the jail source is on tmpfs.

In particular, this avoids modifying the source jail fs when it is not
the true source, as is the case for zfs sources (which use @clean.)

Signed-off-by: Josh Cepek <josh.cepek@usa.net>


bdrewery:
Thanks, merged

PR #284, opened 2015-03-29
Implemented 'local' method for ports (uses local ports tree)

yurivict:
This is useful to quickly rebuild packages based on the local ports tree, with local modifications and patches.

Two notes:
* You may want to add the previously added method 'none' to src/bin/poudriere.8 and doc/poudriere.8.wiki
* You may want to add something like || err 1 "fail" after 'git pull' operation in src/share/poudriere/ports.sh


bdrewery:
This confuses me and I would expect it to confuse users too. 'local' to me means https://github.com/freebsd/poudriere/wiki/use_system_ports_tree but this method _copies_ the ports tree specified. What is the benefit over the existing way to import a tree? If anything I would like to add a `-m null` to replace the obscure documented procedure in that wiki to behave like the `jail -c -m null` feature.

PR #286, opened 2015-03-30
Made colors of status message light gray when corresponding counts are zero

yurivict:


src/share/poudriere/common.sh
```diff
@@ -764,6 +764,14 @@ show_log_info() {
 	return 0
 }
 
+show_colored_count() {
+	if [ $4 -gt 0 ]; then
+		printf "${1}${2}: %-${3}d" ${4}
+	else
+		printf "${COLOR_LIGHT_GRAY}${2}: %-${3}d" ${4}
```
bdrewery, line 771: I think this should be `COLOR_COUNT_ZERO` or `COLOR_DIMMED` or something and then default it to `COLOR_LIGHT_GRAY` in src/share/poudriere/include/colors.sh. That's so a user can change the color without changing what "gray" means.

bdrewery (changes requested):

Issue #289, opened 2015-04-06
poudriere doesn't check ports names for periods


JamesKoenig:
While poudriere checks the jailname string you pass it (with poudriere jail -c) for periods, it doesn't check port names (and likely set names), when running poudriere ports -c, for periods. 

This is a bug because build jails are made with the following pattern: (jailname)-(portname)[-(setname)].  However, jails (man 8 jail) don't allow for periods in their names, though poudriere tries to make one if the <portname> contains periods.

bdrewery:
Related to #247

PR #290, opened 2015-04-06
fix for issue created by allowing periods in port tree name

JamesKoenig:
Fixes a bug which causes poudriere to try and create a jail with a period in it

Sponsored By:  Norse Corp


JamesKoenig:
this is a fix for https://github.com/freebsd/poudriere/issues/289

PR #298, opened 2015-04-21
Implement hooks for ports, adjust sample syntax for www/apache24

ehaupt:
Implement some hooks to be used after creation/updating of a ports tree and provide a sample script that makes sure the updated/created ports tree is always pristine and hasn't been tampered with or hasn't been modified otherwise. This is done by maintaining zfs filesystem snapshots of the created/updated ports tree.

### DIFF
--- a/src/share/poudriere/ports.sh
+++ b/src/share/poudriere/ports.sh
@@ -188,10 +188,14 @@ check_portsnap_interactive() {
 
 if [ ${CREATE} -eq 1 ]; then
 	# test if it already exists
-	porttree_exists ${PTNAME} && err 2 "The ports tree, ${PTNAME}, already exists"
-	maybe_run_queued "${saved_argv}"
 	: ${PTMNT="${BASEFS:=/usr/local${ZROOTFS}}/ports/${PTNAME}"}
 	: ${PTFS="${ZPOOL}${ZROOTFS}/ports/${PTNAME}"}
+	porttree_exists ${PTNAME}
+	if [ $? -eq 0 ]; then
+		run_hook ports create_done ${PTMNT} fail
+		err 2 "The ports tree, ${PTNAME}, already exists"
+	fi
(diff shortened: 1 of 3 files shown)


src/share/poudriere/ports.sh
```diff
@@ -188,10 +188,14 @@ check_portsnap_interactive() {
 
 if [ ${CREATE} -eq 1 ]; then
 	# test if it already exists
-	porttree_exists ${PTNAME} && err 2 "The ports tree, ${PTNAME}, already exists"
-	maybe_run_queued "${saved_argv}"
 	: ${PTMNT="${BASEFS:=/usr/local${ZROOTFS}}/ports/${PTNAME}"}
 	: ${PTFS="${ZPOOL}${ZROOTFS}/ports/${PTNAME}"}
+	porttree_exists ${PTNAME}
+	if [ $? -eq 0 ]; then
```
bdrewery: This won't work right since we use `set -e`. A non-0 return from  `porttree_exists` would exit the script immediately (via the traps we have). It needs to use `if porttree_exists ${PTNAME}; then`

bdrewery: Granted we don't set `set -e` for the `ports` command, we may in the future so it would be better to test the call directly.

src/share/poudriere/ports.sh
```diff
@@ -188,10 +188,14 @@ check_portsnap_interactive() {
...
+		run_hook ports create_done ${PTMNT} fail
```
bdrewery: For hook names the pattern is more "hookname" "event", so these should just be "create fail", "create success", "update start", "update stop".

The `run_hook` command does export  some variables, but not _PTMNT_, so it is normal to pass this as the next argument.

https://github.com/freebsd/poudriere/wiki/hooks

See the others for examples, such as the "jail start" one.

PR #301, opened 2015-04-25
xhtml support

erwinlem:
When the poudriere files are served as xhtml (with the mime type application/xhtml+xml) some errors would occur. These are fixed with these 2 patches. Tested in Safari & Firefox on a mac.

### DIFF
--- a/src/share/poudriere/html/assets/poudriere.js
+++ b/src/share/poudriere/html/assets/poudriere.js
@@ -268,7 +268,7 @@ function build_url(mastername, buildname) {
 		return '';
 	}
 	return 'build.html?' +
-		'mastername=' + encodeURIComponent(mastername) + '&' +
+		'mastername=' + encodeURIComponent(mastername) + '&amp;' +
 		'build=' + encodeURIComponent(buildname);
 }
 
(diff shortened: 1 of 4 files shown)


bdrewery:
Why did you reindent everything? I can't take this as it seemingly is a mega commit with multiple unauditable changes.

I also don't see the reason to use XHTML.

erwinlem:
I reindented because the opening and closing divs didn't match up (there is one too many). The indentation was not correct to begin with (closing body and html tag are at the same level). I could chop it in more fine-grained commits if that is appreciated, but i thought the changes are not that major to begin with. It just looks a lot, it is mostly whitespace.

The big advantage of xhtml is that there is a lot more validation (open close tags must match and stuff like that). Errors are catched early and syntax is forced to be more consistent. Xhtml is also true xml, which means xslt and standard xml tools could be used. I serve my .html files as xhtml by default, so having poudriere support it makes it easier for me. But if xhtml is not wanted i can switch to serving plain html and never think about it again.

mat813:
There's no real advantage, xhtml was an interesting experiment, then came HTML5, which this is, HTML has never been XML.

erwinlem:
xhtml5 is part of the spec, the subtitle of the document is even "A vocabulary and associated APIs for HTML and XHTML". No need to speak in the past tense. http://www.w3.org/TR/html5/introduction.html#html-vs-xhtml

But if none wants to merge i will remove the pull request. At least fix the incorrect open and closing divs yourself. You will run into trouble later on.

bdrewery:
All but the XHTML-specific changes taken. Thanks!

PR #312, opened 2015-05-18
Ignore /var/db/freebsd-update/ too.

mat813:
After a few upgrades, it can end up containing north of a hundred
thousand files, which will slow mtree a lot.


bdrewery:
I blocked even copying this data into the ref and builder jails a few days ago in 1b07fbf484c9bd1a. It saves startup time now as well.

bdrewery:
Well my change only improves TMPFS and UFS usage, not ZFS. I'll still commit your patch.

Issue #259, opened 2014-12-15
Provide support for passing build flags down the dependency chain


marcusva:
# Abstract

poudriere creates packages from ports in isolated jails, which feature their own set of environment variables that are usually set on starting the jail. A dependency X of a port Y thus is unaware of any environment settings the ports framework does on Y's behalf.
In some situations, it is necessary that environment settings are passed down the build chain to allow dependencies to build correctly based on whatever the framework decides "to be the best" for port Y.
# Problem Space

As of writing, the FreeBSD ports tree allows users and ports to build, install and use different versions of the Python language (`lang/python`) at the same time. Ports using the Python bits of the ports framework (namely `Mk/Uses/python.mk`) are enabled to choose a specific Python version or can be marked to support a series of Python versions, giving users (and other ports) the opportunity and flexibility to pick their specific preferred and supported version of the Python language.

The pre-pkg era did not allow ports to install packages for different Python versions, which used the same origin information, which was resolved with the availability of pkg 1.3. The ports and pkg framework are capable of installing e.g. py33-foo _and_ py27-foo originating from `devel/py-foo` at the same time.

A single poudriere instance however is unable to create such a set of packages. Within a single run, it can only create a package py33-foo _or_ py27-foo, but not both. This in turn limits the possibilities of creating different version-specific packages based on the same origin and won't allow different ports to refer to a shared dependency, which _would_ support those versions.
## Previous Workarounds

The limiting situation was resolved in the pre-pkg era by duplicating ports, renaming them and referring to either one of them within the specific consumers.

A port `py-onlypy33` may be limited to Python 3.3 and may depend on `py-foo`, which supports not only Python 3.3, but also Python 2.7. Another port `py-onlypy27`may also depend on `py-foo`, but only works with Python 2.7. This lead to the typical following workaround:
- Create port for py-foo, using Python 2.7
- Create port for py33-foo, using Python 3.3
- py-onlypy33 refers to py33-foo in the `BUILD_DEPENDS` or `RUN_DEPENDS`
- py-onlypy27 refers to py-foo in the `BUILD_DEPENDS` or `RUN_DEPENDS`
## Current Situation within the Ports

[...]

PR #314, opened 2015-05-19
Implemented proper handling of ports with generations (like pyNN-xxx)

yurivict:
This needs to be reviewed and tested. It works for my original purpose.

The additional improvement is that now one can build the python port for non-default version, like this: net/py34-netifaces

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -30,6 +30,8 @@ BSDPLATFORM=`uname -s | tr '[:upper:]' '[:lower:]'`
 . ${SCRIPTPREFIX}/include/common.sh.${BSDPLATFORM}
 BLACKLIST=""
 
+GEN_PREFIXES="py|ruby"
+
 # Return true if ran from bulk/testport, ie not daemon/status/jail
 was_a_bulk_run() {
 	[ "${SCRIPTPATH##*/}" = "bulk.sh" -o "${SCRIPTPATH##*/}" \
@@ -2040,15 +2042,16 @@ gather_distfiles() {
 # Build+test port and return 1 on first failure
 # Return 2 on test failure if PORTTESTING_FATAL=no
 _real_build_port() {
-	[ $# -ne 1 ] && eargs _real_build_port portdir
+	[ $# -ne 1 -a $# -ne 2] && eargs _real_build_port portdir
 	local portdir=$1
 	local port=${portdir##/usr/ports/}
+	local gen="$2"
 	local mnt
 	local log
 	local listfilecmd network
 	local hangstatus
 	local pkgenv phaseenv jpkg
-	local no_stage=$(injail make -C ${portdir} -VNO_STAGE)
+	local no_stage=$(injail_port_make ${port} "${gen}" -VNO_STAGE)
 	local targets install_order
 	local jailuser
 	local testfailure=0
@@ -2075,7 +2078,7 @@ _real_build_port() {
 	else
 		jailuser=root
 		if [ "${BUILD_AS_NON_ROOT}" = "yes" ] &&
-		    [ -z "$(injail make -C ${portdir} -VNEED_ROOT)" ]; then
+		    [ -z "$(injail_port_make ${port} "${gen}" -VNEED_ROOT)" ]; then
 			jailuser=${PORTBUILD_USER}
 			chown -R ${jailuser} ${mnt}/wrkdirs
 		fi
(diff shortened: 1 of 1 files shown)


bdrewery:
Poudriere is not the right place for this, and the current idea of 'generations' is far too specific to python ports and the pkgname. It's an interesting idea but it has to be expanded to the bigger picture of subpackages and multiple packages per origin. Currently I won't take this as it's the wrong place to implement any of this. The framework needs to be mostly self-contained. The feature added in ports is just wrong.

src/share/poudriere/common.sh
```diff
@@ -3105,16 +3126,61 @@ prefix_output() {
...
+}
+
+make_env_gen() {
+	local port="$1"
+	local gen="$2"
+	[ -z "${gen}" ] && return
+	case $(genport_extract_prefix $1) in
+		py) echo "PYTHON_VERSION=$(echo $gen | sed -E "s/(.)(.)/\1.\2/g")"
+		;;
+		ruby) echo "RUBY_VERSION=$gen"
```
bdrewery: This is fundamentally wrong. Poudriere should not have such specific handling here.

bdrewery:
What we do need is to stop having Poudriere only track origins but to track pkgnames and relate those back to origins. We need a bigger picture solution to handle sub-packages and multi-version ports. All of which equate to Many Packages to 1 origin. I think you may actually be on the right track, but I don't like how specific to python this is.

yurivict:
This is just a draft, because I hit this problem for the new port I needed. This of course needs to be expanded and generalized.

Port should return the set of it's generalization classes (make -VCLASSES), framework should know what values are allowed for each class. The idea, and 90% of this is expandable for this. You just picked the piece where python were mentioned. I just couldn't go the whole length in 6 hours that I spent.

bdrewery:
What I would like to see is that a port defines some variable, such as your `CLASSES` idea or `GENERATIONS`, whatever. During the dependency lookup in `list_deps` we lookup this variable. If it has values then we act on those to generate virtual packages based on them. This may inject extra dependencies into the queue, such as in the python case to pass along this `DEPENDS_ARGS` flag (which I'd like to see changed). Poudriere would not need to be so strict about unique origins anymore. It should just build every package with the parameters stored with that package in the queue. When it builds it should just pass along `GENERATION=whatever` and have the framework handle the `DEPENDS_ARGS` etc.

yurivict:
I will generalize this the way you suggested. The additional ideas are:
* Some serial port options (like DOCS, DEBUG, SOUND), or maybe even all port options can become "classes"
* Before the build, user can set the list of default class values (ex. DOCS=no, DOXYGEN=no) to the build tool (poudriere in this case), and such choices will be made for corresponding classes when they are available, but undefined. Choosing PYTHON=3.4 will effectively override the default python version.
* package names will reflect classes. For example, docsY-debugY-py31-netifaces may refer to the package for the port origin net/netifaces with DOCS=yes, DEBUG=yes, python version 31. (like this is now)
* class choices will be passed to the port make through the environment variable PORT_CLASSES. (instead of current PYTHON_VERSION)
* port infrastructure based on the class selection and its own logic will decide what dependencies it will have, and what classes dependencies will have. It will will return the list with class choices. (currently this is deduced from prefixNN- syntax for dependency). XXX_DEPENDS in port makefiles need to be generalized.

The current version implies one "class", "PYTHON", when the port name begins with py-, and no classes otherwise.

bdrewery:
I keep making slave ports for my own needs. For example I have `mail/postfix` with _DOVECOT_ option. But I want a non-dovecot-depending version as well. So I added a `local/postfix-relay` as a slave port that I can set my own options on. The do conflict but that is fine in my case.

My need combined with the python needs seems like they could be solved in a unified way.

yurivict:
This week I am going to commit the next phase that will add the build class, with many current binary options and python version becoming class parameters, so that poudriere will build with these options. Poudriere itself will only have the machinery, and no knowledge of particular classes. Yet next iteration should be concerned with how to pass class requirements  into dependencies in a nice way.

koobs:
@yurivict Thank you for taking the initiative and your effort in working on this. It's probably worth introducing yourself to the rest of the FreeBSD / Python team on IRC (#freebsd-python @ freenode) so that we can help make progress :) cc @marcusva

Also see: #259

bdrewery:
Thank you for your PR and patch to support it.  Since you sent it in though the dependency calculation code has been rewritten (last year) to improve parallelism.  Thus the patch was 70% no longer applicable. 

I implemented support for _DEPENDS_ARGS_ in #483, as well as _FLAVORS_.

PR #316, opened 2015-05-21
Fix test syntax for string

rbgarga:
test uses = for string comparison, -eq is for integer


bdrewery:
I wonder what the point of setting WITH_ELFTOOLCHAIN_TOOLS is. It's only supposed to change which toolset is used.

bdrewery:
The answer is that the GNU binutils strip(1) doesn't work on mips.

rbgarga:
According UPDATING, this is the default since January, then this can be removed, no?

20150107:
	ELF tools addr2line, elfcopy (strip), nm, size, and strings are now
	taken from the ELF Tool Chain project rather than GNU binutils. They
	should be drop-in replacements, with the addition of arm64 support.
	The WITHOUT_ELFTOOLCHAIN_TOOLS= knob may be used to obtain the
	binutils tools, if necessary.

PR #317, opened 2015-05-21
Fix jail update with patch:

rbgarga:
Revert previously applied patch and remove orphan files before attempt
to run 'svn up'. Otherwise conflicts can break src.

Reapply patch after update src.


bdrewery:
I'm still a bit nervous about putting this change in. I think we might cause unexpected problems by always reverting before an update. Someone might purposely have their own customizations in the tree that they could lose here silently.

Plus it might make more sense to generalize this and have the entire jail rolled back to the "clean" snapshot before upgrade.

PR #322, opened 2015-06-21
New poudriere.conf ZJAIL and ZPORTS

pgollucci:


src/share/poudriere/jail.sh
```diff
@@ -617,6 +617,7 @@ create_jail() {
 
 	if [ -z "${JAILFS}" -a -z "${NO_ZFS}" ]; then
 		[ -z ${ZPOOL} ] && err 1 "Please provide a ZPOOL variable in your poudriere.conf"
+                [ -n ${ZJAIL} ] && ZPOOL=$ZJAIL
```
bdrewery: This should come 1 line earlier.

src/etc/poudriere.conf.sample
```diff
@@ -7,10 +7,23 @@
 # poudriere will use tank/${ZROOTFS} as its root
 #
 # You need at least 7GB of free space in this pool to have a working
-# poudriere.
+# poudriere. 
 #
 #ZPOOL=tank
 
+# You may also optionally define ZJAIL and/or ZPORTS.  If unset they
```
bdrewery: I think we're already too complex with ZPOOL, ZROOTFS, BASEFS and POUDRIERE_DATA, adding 2 more variables in doesn't sound good to me.

Just looking at the actual code change I don't see how it helps, which means other code before it is now obscured by the ZJAIL/ZPORTS variables.

The POUDRIERE_DATA var is already in the sample config and overridable. I think setting the zfs mountpoint can achieve doing what you want without another variable.

PR #323, opened 2015-06-21
Update .gitnores

pgollucci:


.gitignore
```diff
@@ -27,12 +27,17 @@
...
 config.log
 host-upgrade.sh
+libnv.a
+libucl.a
 poudriere.fossil
 poudriered
 src/bin/poudriere.8.gz
 src/config.h
+src/etc/poudriere.conf
+src/etc/poudriere.d
```
bapt: This one sounds wrong we have samples in there in the sources

pgollucci:
fixed.  Good catch.  Thanks.

.gitignore
```diff
@@ -27,12 +27,18 @@
...
 config.log
 host-upgrade.sh
+libnv.a
+libucl.a
 poudriere.fossil
 poudriered
 src/bin/poudriere.8.gz
 src/config.h
+src/etc/poudriere.conf
+src/etc/poudriere.d/hooks
```
bdrewery: We have valid files checked into this directory.

It's better to just use -e to not use the source's etc dir.

PR #332, opened 2015-07-16
Make it possible to use jail -c -a arch -m src=/usr/src when arch is …

kpaasial:
…not equal to host uname -m and world has been cross-compiled using /usr/src sources and TARGET set to arch.

This fix should allow for example on a amd64 host:

poudriere jail -c -a i386 -v stable/10 -m src=/usr/src -j jailname

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -405,6 +405,9 @@ build_and_install_world() {
 install_from_src() {
 	local cpignore_flag cpignore
 
+	export TARGET=${ARCH%.*}
+	export TARGET_ARCH=${ARCH#*.}
+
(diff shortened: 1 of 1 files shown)


bdrewery:
This doesn't seem right. I'm looking.

kpaasial:
I guess more information is in order. Without my fix 'poudriere jail -c -a i386 -v stable/10 -m src=/usr/src -j jailname' creates the jail but since TARGET/TARGET_ARCH are not set to i386 'make installworld' installs an amd64 jail. I have verified this by looking at what gets installed in the jail.

bdrewery:
I believe you, I just think we have code that should be covering this already that is likely not connected right.

bdrewery:
Fixed!

PR #342, opened 2015-08-24
Create hiawatha.conf.samples

bsdjunk:
Add in sample for hiawatha


src/share/examples/poudriere/hiawatha.conf.samples
```diff
@@ -0,0 +1,12 @@
+Please copy this to your hiwatha.conf
```
bdrewery: Dumb question but does Hiawatha allow comments anywhere like this? It's also spelled wrong.

bapt:
feedback timeout

Issue #341, opened 2015-08-16
What about "~/ports-cvs/mybeautifulporttotest"?


uzsolt:
I found it in the manual (EXAMPLES/test a single port):

     It starts the jail, then mount the ports tree (nullfs), then mounts the
     package dir (pourdriere/data/packages/<jailname>-<tree>-<setname>), then
     it mounts the ~/ports-cvs/mybeautifulporttotest (nullfs) it builds all
     the dependencies (except runtime ones) and log it to
     poudriere/data/logs/testport/jailname/default/mybeautifulporttotest.log).

I tried, but the ~/ports-cvs doesn't mount as the manual describe - and didn't find "ports-cvs" in source files.

Is it a bug in manual or am I missing something?

f-andrey:
If I understand correctly, the relics of the old documents are no longer current.
I have prepared a patch
https://github.com/freebsd/poudriere/compare/master...f-andrey:master

PR #349, opened 2015-09-11
Fix issue #341

f-andrey:
Fix issue #341 
poudriere.8 and doc/poudriere.8.wiki contain relics information, this patch fix this and update examples for uniformity


bdrewery:
Btw there's no need to update the .wiki file as it is generated.

bdrewery:
I appreciate this, but there's a ton of grammar bugs around in still in the changed lines. It would also be better to not hardcode "default" but rather use something like "${PORTSTREE}" and "${SETNAME}" and for the time format something like "${BUILDNAME_FORMAT}" since that is the variable name in the config.

PR #365, opened 2015-11-27
FreeBSD 9 rm doesn't have -x using find's -x and -delete

derekschrock:


brd:
I wonder if it is worth making this conditional on what version of the OS, so we can clean it back up later.

@bdrewery Thoughts?

bdrewery:
I would rather stick to using rm. The -x flag can be handled in our rm() function based on the osversion.

derekschrock:
How do you want to handle the -x in rm()?

If we're <10 remove x from all flag-like args and not worry about cross mount rm in <10?
If we're <10 and using x use 9's find -x to mimic 10's rm -x?

If the latter are we going to assume "$2" is the directory to be deleted or do we need to use all positional parameters after the last arg/--.

bdrewery:
Fix committed to ports for now (revert).

bdrewery:
Fixed by a0c66afa62e91da758a2394ac52ce626c289d5d2 and ad93bf21f8b28b838320f927e67446773f0a4e0e

PR #380, opened 2016-03-11
Move XDEV check into installworld()  so that -m src will do the right…

seanbruno:
… thing when -x is passed.

This also respects SRC_DIR now so that -m src= actually works with -x instead of grabbing /usr/src versions of native-xtools.


bdrewery:
I'll review this in the next few weeks. Quite busy at the moment.

bdrewery:
Using SRC_BASE reverts my change to always use /usr/src.  It's not fixed such that this is safe to change back yet.  native-xtools is only safe to build with host srcs since it has no sysrooting, it is using all of the host libraries and headers.

seanbruno:
Abandon.  Will come back at this when we can make native-xtools use sysroot or something sane.

Issue #291, opened 2015-04-09
regression-test target support


AMDmi3:
We need subj. Should as easy as running an extra target near where stage-qa is run, and should be subject to the same bulk -t / -r logic.

bdrewery:
Duplicate of #355

PR #382, opened 2016-03-30
Run test target on testport and bulk -t

grembo:
This is an attempt to address #291

It might make more sense to provide this as a separate
command line option though.

It would also be good to have a way to automatically set
option TEST (should be optional though, as you want
to be able to test ports without this option set)

Also, max execution time is a bit arbitrary.


grembo:
I guess this is kind of handled by https://github.com/freebsd/poudriere/pull/355 already?! Should've checked first. Can we get that in soon?

bdrewery:
Duplicate of #355

PR #383, opened 2016-04-11
Switch BUILD_AS_NON_ROOT as default to on.

mat813:
It's been long enough, there should be no ports that don't build as root, and I'm tired of going over them every month and fixing them...

### DIFF
--- a/src/etc/poudriere.conf.sample
+++ b/src/etc/poudriere.conf.sample
@@ -221,8 +221,8 @@ DISTFILES_CACHE=/usr/ports/distfiles
 #PRESERVE_TIMESTAMP=yes
 
 # Define to yes to build and stage as a regular user
-# Default: no
-#BUILD_AS_NON_ROOT=yes
+# Default: yes
+#BUILD_AS_NON_ROOT=no
 
 # Define pkgname globs to boost priority for
 # Default: none
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -4762,7 +4762,7 @@ fi
 : ${PIDFILE:=${POUDRIERE_DATA}/daemon.pid}
 : ${QUEUE_SOCKET:=/var/run/poudriered.sock}
 : ${PORTBUILD_USER:=nobody}
-: ${BUILD_AS_NON_ROOT:=no}
+: ${BUILD_AS_NON_ROOT:=yes}
 : ${DISTFILES_CACHE:=/nonexistent}
 : ${SVN_CMD:=$(which svn 2>/dev/null || which svnlite 2>/dev/null)}
 # 24 hours for 1 command


bdrewery:
Yup. Need to add some logic about ccache since they don't work together. I'll handle that.

Regards,
Bryan Drewery

mat813:
Mmm, I've been using both for a long time, and do work just right together.

AMDmi3:
> Mmm, I've been using both for a long time, and do work just right together.

Same here.

mat813:
To be fair, there is one port that does not build as a user, it is devel/spark, I have not fixed it because I don't know 1) java 2) scala 3) maven, and don't want to :)
But I'm sure the fix is easy for someone who knows it. (And I really don't feel it is a blocker)

bdrewery:
There's no way it can work by default since the CCACHE_DIR is going to be default root owned and building as user is not root, it is nobody. Nobody cannot write into the CCACHE_DIR. You must chmod a+w in the CCACHE_DIR at least.

bdrewery:
The chmod also creates a security problem.

AMDmi3:
Using ```nobody``` is also not the best solution from security standpoint. Better to use dedicated user.

mat813:
A dedicated user would mean that a user need to be added in the jail, which is not really a good idea, nobody is perfect for that.

AMDmi3:
> A dedicated user would mean that a user need to be added in the jail, which is not really a good idea

I don't see any problems with that. We already add resolv.conf, make.conf and other stuff on jail startup, adding a user is just one more command.

 > nobody is perfect for that

```nobody``` is actually not good for anything and should never be used. The fundamental problem with ```nobody``` is that more than a single service may run under it. The more services are run as ```nobody```, the more privileged it becomes, ending up as another superuser.

What that basically means is that when poudriere ccache cache is owned by ```nobody```, and any other service running as ```nobody``` is compromised, cache would get poisoned, malicious code would end up in the packages and all machines which fetch from that repository would be rooted.

bdrewery:
We do already add the `PORTBUILD_USER` in the jail if not present. That's not a problem. It is added as uid _65532_.

What is a problem is POLA and breaking Poudriere for users.  Sure you could setup ccache in the right configuration to build most posts with non-root, but not all.  Most Poudriere users don't know how or care to debug the obscure failures that come from broken ccache+non-root.  So it must just work by default.  Sharing a ccache dir can't work between Poudriere and root, which because of how Poudriere uses it, `CCACHE_DIR` vs `CCACHE=yes`, is going to be pretty common since the user is providing a directory that is likely being shared elsewhere.

What's still lacking is any documentation anywhere explaining how to do `BUILD_AS_NON_ROOT` and ccache together.  Until that time I will keep the default off `BUILD_AS_NON_ROOT` when using ccache.  I will add an override though.

bdrewery:
I'm going to make the default `no` with `CCACHE_DIR` and give a warning when both are enabled.

AMDmi3:
User does need to do anything or care of anything, poudriere can and should do everything itself - create private directory for its ccache and properly chown it. Another directory for root's ccache, if you still want to support it. I don't understand what's the problem with this.

bdrewery:
I'm repeating myself now.  Poudriere does not own or create the ccache dir, it is provided by the user.  Because the user provides it then Poudriere cannot make assumptions about what it may do to it.  It was never documented like that and it would be a POLA violation to go messing with it.

AMDmi3:
> I'm repeating myself now. Poudriere does not own or create the ccache dir, it is provided by the user

I'm repeating myself as well, poudriere must create and own it's own ccache dir, and use it for buids. User is free to override it with his own directory.

bdrewery:
Feel free to write your own tool.  FYI I've added you to my ignore list.

PR #401, opened 2016-07-01
Fix BASEFS over NFS if the NFS server's TZ differed from the local TZ

kdkd:
If BASEFS is on an NFS mount, nohang was immediately killing builds
if the NFS server's timezone differed from the local timezone. This
appears to be because fifos created over NFS initially have their
mtime set to the server's local time not the clients. Working around
this by adding a "touch" after creating the nohang fifo.

### DIFF
--- a/src/share/poudriere/include/parallel.sh
+++ b/src/share/poudriere/include/parallel.sh
@@ -250,6 +250,9 @@ nohang() {
 
 	fifo=$(mktemp -ut nohang.pipe)
 	mkfifo ${fifo}
+	# If the fifo is over NFS, newly created fifos have the server's
+	# mtime not the client's mtime until the client writes to it
+	touch ${fifo}
 	exec 8<> ${fifo}
 	rm -f ${fifo}
 


bdrewery:
I was just documenting this for a release and am confused now.  I don't see the fifo file's mtime being used by nohang at all so I'm not sure how this achieves anything.  A few lines down there is a `stat -f` done on the *logfile*, which if it were on NFS could cause the problem reported here I think.

bdrewery:
```
# nohang()

lastupdated=$(stat -f "%m" ${logfile})              
now=$(date +%s)                                 
                                                    
# No need to actually kill anything as stop_build() 
# will be called and kill -9 -1 the jail later      
if [ $((now - lastupdated)) -gt $log_timeout ]; then
        ret=2                                       
        break                                       
elif [ $((now - starttime)) -gt $cmd_timeout ]; then
        ret=3                                       
        break                                       
fi                                                  
```

In this, the `log_timeout` is `NOHANG_TIME` (7200).

bdrewery:
Ok I think I see the problem now.  This fix was only partial and hid the other problem.

First the `read -t` below times out.  At 7200 seconds this timeout is set to 720 seconds.  So fixing the fifo timestamp allows the read to go for 720 seconds before it loops around and does the `stat -f` comparison.  At an instant `read -t` timeout due to the fifo mtime, it finds the logfile mtime is wildly different than the local server time and bombs out.  So the fix committed here appears to only add 720 seconds before hitting the `stat -f` comparison.  `nohang` is ran *per-phase*, so as long as each build phase is under 720 seconds there is no problem.

The one thing I can't explain is that since `stat -f` is done before `read -t`, how it doesn't immediately cause a problem and exit before hitting the timeout.  This is of course working on the assumption that the patch here fixed the problem for you.  If so I guess there may be some other kernel caching occurring which allows the first `stat -f` comparison to succeed.  I am more inclined to think that it is just immediately hitting the error and that touching the fifo is probably needed but not enough.

Here's the relevant part of the function (before the touch) with 2 `msg` added in for debug:

```
read_timeout=$((log_timeout / 10))

fifo=$(mktemp -ut nohang.pipe)
mkfifo ${fifo}
exec 8<> ${fifo}
rm -f ${fifo}

starttime=$(date +%s)

# Run the actual command in a child subshell
(
        local ret=0
        _spawn_wrapper "$@" || ret=1
        # Notify the pipe the command is done
        echo done >&8 2>/dev/null || :
        exit $ret
) &
childpid=$!
echo "$childpid" > ${pidfile}

# Now wait on the cmd with a timeout on the log's mtime
while :; do
        if ! kill -CHLD $childpid 2>/dev/null; then
                _wait $childpid || ret=1
                break
        fi

        lastupdated=$(stat -f "%m" ${logfile})
        now=$(date +%s)

        # No need to actually kill anything as stop_build()
        # will be called and kill -9 -1 the jail later
        if [ $((now - lastupdated)) -gt $log_timeout ]; then
                ret=2
                break
        elif [ $((now - starttime)) -gt $cmd_timeout ]; then
                ret=3
                break
        fi

        # Wait until it is done, but check on it every so often
        # This is done instead of a 'sleep' as it should recognize
        # the command has completed right away instead of waiting
        # on the 'sleep' to finish

        msg "WAITING"

        unset n; until trappedinfo=; read -t $read_timeout n <&8 ||
                [ -z "$trappedinfo" ]; do :; done

        msg "TIMEOUT or done"

        if [ "${n}" = "done" ]; then
                _wait $childpid || ret=1
                break
        fi
        # Not done, was a timeout, check the log time
done
```

```
=======================<phase: check-sanity   >============================
====>> WAITING
/!\ WARNING /!\
Ports Collection support for your FreeBSD version has ended, and no ports are guaranteed to build on this system. Please upgrade to a supported release.

===>   NOTICE:

The cvs port currently does not have a maintainer. As a result, it is
more likely to have unresolved issues, not be up-to-date, or even be removed in
the future. To volunteer to maintain this port, please create an issue at:

https://bugs.freebsd.org/bugzilla

More information about port maintainership is available at:

https://www.freebsd.org/doc/en/articles/contributing/ports-contributing.html#maintain-port

===>  License GPLv2+ accepted by the user
====>> TIMEOUT or done
===========================================================================
=======================<phase: pkg-depends    >============================
===>   cvs-1.20120905_4 depends on file: /usr/local/sbin/pkg - not found
===>   Installing existing package /packages/All/pkg-1.9.4_1.txz
[exp-10amd64-commit-test] Installing pkg-1.9.4_1...
[exp-10amd64-commit-test] Extracting pkg-1.9.4_1: .......... done
===>   cvs-1.20120905_4 depends on file: /usr/local/sbin/pkg - found
===>   Returning to build of cvs-1.20120905_4
===========================================================================
=======================<phase: fetch-depends  >============================
===========================================================================
=======================<phase: fetch          >============================
====>> WAITING
```

So you can see that the WAITING->"TIMEOUT or done" covers the entire phase.

I am going to swap the ordering here to do the `read` first rather than the `stat -f`, as in most cases that should avoid ever running `stat -f`.

I don't have a setup to test any of this over NFS, if you could double check that this patch did fix the issue for you and that reverting it does restore the issue, then I have some idea for debugging further.

PR #403, opened 2016-07-13
Add small note on how to disable tmpfs(5) usage

t6:
To disable tmpfs(5) in Poudriere, I first tried `USE_TMPFS=no` because of the similar `USE_PORTLINT=no` above it, but that counter-intuitively didn't work. Commenting `USE_TMPFS` helped.

I hope adding a small note about this to `poudriere.conf.example` will make this a little clearer.


bdrewery:
Hmm, I was going to suggest `no` be documented instead, but if it is broken it needs to be fixed too.

t6:
Ok, I amended the patch to do that instead.

I tried to preserve the existing behavior i.e. commenting `USE_TMPFS` means the same as `USE_TMPFS=no`.

bdrewery:
I simplified it slightly in 1c03d8d51f1b0e382da95dea86573881cd9ce324

PR #404, opened 2016-07-13
image: Firmware mode improvement

ocochard:
- Add nanobsd diskless(8) mode.
  firmware use read-only root filesystem, then it needs to use ram disk for /etc (and /usr/local/etc) and /var
  nanobsd use a very specific diskless(8) mode for allowing to copy /etc (and /usr/local/etc) into the specific cfg partition
  This backup script can be the nanobsd one (/usr/src/tools/tools/nanobsd/Files/root/save_cfg) or others nanobsd based project script
- Replace root filesystem gpt label from 3-4 to 1-2 (like nanobsd)
- Exclude some big useless directories to be installed on the image like var/db/freebsd-update, var/db/etcupdate and boot/kernel.old
  This folder are left after poudriere jail update and useless in a firmware case
- Remove the SUJ (then the big .journal file) because as read-only filesystem, it's usage is useless too

There is still one bug to fix, but I don't find the root cause:
On a poudriere-firmware image running, permission of /etc (a ramdisk created by /etc/rc.initdiskless) are wrong.
Replacing usage of tar by cpio didn't fix the problem, I need to check if replacing usage of makefs/mkimg by old mdconfig fix this strange permission problem.

### DIFF
--- a/src/share/poudriere/image.sh
+++ b/src/share/poudriere/image.sh
@@ -126,6 +126,11 @@ post_getopts
 : ${IMAGENAME:=poudriereimage}
 MASTERNAME=${JAILNAME}-${PTNAME}${SETNAME:+-${SETNAME}}
 
+# CFG_SIZE set /etc and /var ramdisk size and /cfg partition size
+# DATA_SIZE set /data partition size
+CFG_SIZE="32m"
+DATA_SIZE="32m"
+
 # Limitation on isos
 case "${IMAGENAME}" in
 ''|*[!A-Za-z0-9]*)
@@ -155,14 +160,19 @@ excludelist=$(mktemp -t excludelist)
 mkdir -p ${WRKDIR}/world
 mkdir -p ${WRKDIR}/out
 [ -z "${EXCLUDELIST}" ] || cat ${EXCLUDELIST} > ${excludelist}
+# Some poudriere jail directory used for upgrading source or binary based jail
+# are useless on the final image
 cat >> ${excludelist} << EOF
 usr/src
+var/db/freebsd-update
+var/db/etcupdate
+boot/kernel.old
 EOF
 case "${MEDIATYPE}" in
 usb|*firmware|rawdisk)
 	truncate -s ${IMAGESIZE} ${WRKDIR}/raw.img
 	md=$(/sbin/mdconfig ${WRKDIR}/raw.img)
-	newfs -j -L ${IMAGENAME} /dev/${md}
+	newfs -L ${IMAGENAME} /dev/${md}
 	mount /dev/${md} ${WRKDIR}/world
 	;;
 zrawdisk)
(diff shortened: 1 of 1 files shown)


src/share/poudriere/image.sh
```diff
@@ -155,14 +160,19 @@ excludelist=$(mktemp -t excludelist)
...
+var/db/freebsd-update
+var/db/etcupdate
+boot/kernel.old
 EOF
 case "${MEDIATYPE}" in
 usb|*firmware|rawdisk)
 	truncate -s ${IMAGESIZE} ${WRKDIR}/raw.img
 	md=$(/sbin/mdconfig ${WRKDIR}/raw.img)
-	newfs -j -L ${IMAGENAME} /dev/${md}
+	newfs -L ${IMAGENAME} /dev/${md}
```
bapt: We should find a way to make this change optional so one can decide if he wants journalising or not

bapt:
What do you mean by wrong: owners are wrong? on all files, or permissions are wrong?

src/share/poudriere/image.sh
```diff
@@ -262,11 +272,51 @@ usb|rawdisk)
...
+        mkdir -p ${WRKDIR}/world/etc/local
+		tar -C ${WRKDIR}/world -X ${excludelist} -cf - usr/local/etc/ | tar -xf - -C ${WRKDIR}/world/etc/local
+        rm -rf ${WRKDIR}/world/usr/local/etc
+        ln -s /etc/local ${WRKDIR}/world/usr/local/etc
+    fi
+
+	# Copy /etc and /var to /conf/base as "reference"
+	for d in var etc; do
+		mkdir -p ${WRKDIR}/world/conf/base/$d ${WRKDIR}/world/conf/default/$d
+		tar -C ${WRKDIR}/world -X ${excludelist} -cf - $d | tar -xf - -C ${WRKDIR}/world/conf/base
```
bapt: maybe add the -p argument here in both tar commands?

ocochard: Once poudriere-firmware boot, the resulting system show this kind of error message:
`sendmail: can not open config '/usr/local/etc/dma/dma.conf': Permission denied
Jul 14 12:00:00 router cron[2673]: _secure_path: cannot stat /etc/login.conf: Permission denied`

Then I've check the /etc permission:
```
[root@poudriere-img]~# ll / | grep etc
d---------  25 root  wheel     2560 Jul 13 18:23 etc/
```
/etc is a ram disk created by diskless(8) (like on nanobsd):
```
[root@poudriere-img]~# mount
/dev/gpt/BSDRP1 on / (ufs, local, read-only)
devfs on /dev (devfs, local, multilabel)
/dev/md0 on /etc (ufs, local)
/dev/md1 on /var (ufs, local)
```
On a standard FreeBSD and nanobsd permission of /etc should be:
```
[root@nanobsd]~# ll / | grep etc
drwxr-xr-x  27 root  wheel     2048 Jun 12 05:30 etc/
```


And if I unmount this /etc, the underlying /etc regain it's original (and good) permissions:
```
[root@poudriere-img]~# umount /etc
[root@poudriere-img]~# ll / | grep etc
drwxr-xr-x  24 root  wheel     2560 Jul 11 16:27 etc/
```

This is why I've no clue about where the permission problem is inserted during the poudriere image build.
About adding "-p" to the 2 tar commands: I've just tried it, but it didn't solve the problem.
I will be AFK since now to 8th August.

bdrewery:
Ping. What is needed to move forward with this? I want to do a 3.2 release soon which will be the first for `poudriere image`.

ocochard:
I've wrote up my experience here: https://bsdrp.net/documentation/technical_docs/poudriere
What need to be done in priority:
1. Image size: It need to ask for the final full image size (a little bit trimmed) and not only one system partition (If I'm generating an image for a 2Gb flash disk, I don't want to obtain a 4.2Gb image like currently)
2. Fix the /etc/ that is currently not in a ramdisk: /etc need to be, like /var, into a ramdisk too (and automatically populate with the diskless feature like /var)
3. Once /etc will be a ramdisk, we will need a "save config" script like with nanobsds for copying the diff between the ramdisk /etc and reference /conf/base/etc/ into the /cfg partition. re-using the minimum /usr/src/tools/tools/nanobsd/Files/root/save_cfg script can be enought.

brd:
I just fixed 1) in https://github.com/freebsd/poudriere/commit/14f5567bd763701c5bbd50634c5a6b21458eda7c.

ocochard:
I've found the source of the /etc permission problem into firmware mode: It's related to the default permissions that makefs put on the cfg and data partition image. Will try to found a solution about it.

PR #405, opened 2016-07-28
Add support for file:/// URLs with git

gvnn3:
Adding this line allows poudriere to grab the ports tree from a repo on a local filesystem.


bapt:
Why using git in that case in poudriere and not directly the local filesystem?

gvnn3:
I felt that file:/// was missing and ought to work.

AMDmi3:
> Why using git in that case in poudriere and not directly the local filesystem?

Obviously because it's a different repo which may be worked on independently.

Issue #391, opened 2016-05-19
Dtrace builds


dinapappor:
I am having a difficult time building packages with dtrace which is needed for observability.

I have tried copying /boot/ into the reference jail, both manually and adding it to common.sh operations. 

It's still a no go, how does one go about compiling packages with dtrace supports?

dinapappor:
```
checking for 1-stage DTrace precompilation... dtrace: failed to compile script emulator/beam/erlang_dtrace.d: "/usr/lib/dtrace/ip.d", line 225: operator -> cannot be applied to a forward declaration: no struct ip definition is available
configure: error: Could not precompile erlang_dtrace.d: dtrace -h failed
configure: error: /bin/sh '/wrkdirs/usr/ports/lang/erlang/work/otp_src_18.3/erts/configure' failed for erts
===>  Script "configure" failed unexpectedly.
Please report the problem to olgeni@FreeBSD.org [maintainer] and attach the
"/wrkdirs/usr/ports/lang/erlang/work/otp_src_18.3/config.log" including the
output of the failure of your make command. Also, it might be a good idea to
provide an overview of all packages installed on your system (e.g. a
/usr/local/sbin/pkg-static info -g -Ea).
*** Error code 1

Stop.
make: stopped in /usr/ports/lang/erlang
====>> Cleaning up wrkdir
===>  Cleaning for erlang-18.3.3_1,3
build of lang/erlang ended at Thu May 19 17:01:00 UTC 2016
```

This is what happends during configure phase.

dinapappor:
Actually, copying the /boot/ directory to the poudrier worker jails actually worked. \o/ Is it possible to have an "ENABLE_DTRACE=YES" option that would actually do this?

bdrewery:
What from /boot is needed? We can just make the jail create/update ensure the files are brought in.

dinapappor:
@bdrewery I think it's these files:

```
/boot/kernel/dtrace.ko              /boot/kernel/dtrace.ko.symbols      /boot/kernel/dtrace_test.ko         /boot/kernel/dtrace_test.ko.symbols /boot/kernel/dtraceall.ko           /boot/kernel/dtraceall.ko.symbols
```

I will make some experiments and try to narrow it down to exactly what files are needed and get back to you.

dinapappor:
Hi, I looked this up and this is what I did to be able to build dtrace enabled packages repeatedly:

In common.sh after line 1776 I added: 

```
[...]

PR #410, opened 2016-08-26
Getting dtrace to work under poudriere

infracaninophile:
Ref #931

Seems that there are 3 prerequisites:
- kldload dtraceall
- unhide /dev/dtrace and /dev/dtrace/\* in the build jail
- Access to the kernel and various module binaries via a populated /boot/kernel (possibly also /boot/modules)

This adds code to do all of those three things when 'USE_DTRACE=yes' is set in poudriere.conf


infracaninophile:
Err -- that should say Ref #391.

bdrewery:
thanks!

infracaninophile:
So, I did a bunch of test builds (ie. poudriere testport) of postgresqXY-server with the DTRACE option turned on using a number of different environments.

The good news: these patches work well for building dtrace enabled pkgs in a 9.3 or 10.1 jail on an 11-STABLE host.

The bad news: utter and ignominious failure to build i386 packages (for any OS version) on an amd64 box.

swills:
Adding -xnolibs to calls to dtrace in the ports that fail to build/configure dtrace in jail might be sufficient and easier.

bdrewery:
Is this safe to merge given the cross-build issue? We could just disable some of these changes in cross-build scenarios.

infracaninophile:
Well, it will cause a cross-built port to fail to compile when USE_DTRACE=yes in poudriere.conf if the port is configured to build DTRACE probes.   Otherwise it should be safe.

infracaninophile:
Having spent some quality time with Google, I think @swills comment above is probably the correct approach here.  There are something like 30 ports that have dtrace probes available.  Fixing those to add -xnolibs to the dtrace command line looks like a more productive route forwards.  Some have already been fixed: for instance lang/ruby24 already enables dtrace and defaults to adding -xnolibs automatically.

Closing and withdrawing this pull request

bdrewery:
Wait, isn't swills just suggesting a fix for the cross-build problem? Wouldn't we still want your patches here?  There's been multiple PR for adding dtrace support that I have not merged yet.  Another was #391.  I would really like it to "just work".

infracaninophile:
On 04/06/2017 16:53, Bryan Drewery wrote:
> Wait, isn't swills just suggesting a fix for the cross-build problem? Wouldn't we > still want your patches here?  There's been multiple PR for adding dtrace support > that I have not merged yet.  Another was #391.  I would really like it to "just 
> work".


As I understand it, if you call dtrace as

   /usr/sbin/dtrace -xnolibs ...

then you don't need the dtrace kernel module loaded in order to compile
the dtrace probes.  It should 'just work' (tm) in a standard poudriere
build jail without a kernel present.  Pretty sure that also means it
should work for cross compiling too, although that needs confirmation.

Like I said, if dtrace works for the existing standard ports build of
ruby-2.4.1, where it is unconditionally enabled and has the -xnolibs
flag applied, then I see no reason why this approach shouldn't work
everywhere.

Works for me on amd64:
```
[lucid-nonsense]:...local/etc/poudriere.d:# ruby24
^Z
Suspended
[lucid-nonsense]:...local/etc/poudriere.d:# dtrace -l | grep ruby
69078  ruby23171   libruby24.so.24                       rb_ary_drop
array-create
69079  ruby23171   libruby24.so.24                    rb_ary_product
array-create
69080  ruby23171   libruby24.so.24       rb_ary_repeated_combination
array-create
69081  ruby23171   libruby24.so.24       rb_ary_repeated_permutation
array-create
69082  ruby23171   libruby24.so.24                rb_ary_combination
array-create
69083  ruby23171   libruby24.so.24                rb_ary_permutation
array-create
69084  ruby23171   libruby24.so.24                     rb_ary_sample
array-create
69085  ruby23171   libruby24.so.24                        rb_ary_and
array-create
69086  ruby23171   libruby24.so.24                       rb_ary_diff
array-create
69087  ruby23171   libruby24.so.24                      rb_ary_times
array-create
[...]
```

I don't actually have any machines with a different architecture so I
can't test whether cross-building really does result in working dtrace
bits though.

	Cheers,

	Matthew

Issue #176, opened 2014-09-18
Poudriere options doesn't save options in $jail-$portstree-options when using -p


bapt:
If no ports tree is given with -p it defaults to "default". In this case we need to empty it. 

The following patch could be a first approach: http://pastebin.com/KR1uQbkd

bdrewery:
#583 covers this better (duplicate).

PR #419, opened 2016-10-25
Honoring the -p option in option directory name in the options command

pizzamig:
The directory storing ports options doesn't contain the ports tree name
provided in the command line. It will be added if and only if a port tree
is provided (even if it's default)
This error is already reported by bapt@ (bug #176 ); I slightly improved his solution, ignoring the portstree name in the directory name if no -p is provided, but honoring it if "-p default" is used.

### DIFF
--- a/doc/poudriere.8.wiki
+++ b/doc/poudriere.8.wiki
@@ -762,6 +762,10 @@ It is very easy to extend it so that we can easily add other tests if wanted.</d
 </blockquote>
 <blockquote style="margin-top: 0.00em;margin-bottom: 0.00em;">
 <div class="display">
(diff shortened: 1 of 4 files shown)


doc/poudriere.8.wiki
```diff
@@ -872,6 +884,10 @@ As a starter, you may want to copy an existing /var/db/ports/ to /usr/local/etc/
 </blockquote>
 <blockquote style="margin-top: 0.00em;margin-bottom: 0.00em;">
 <div class="display">
+<code class="lit">/usr/local/etc/poudriere.d/&lt;tree&gt;-&lt;setname&gt;-make.conf</code></div>
+</blockquote>
+<blockquote style="margin-top: 0.00em;margin-bottom: 0.00em;">
+<div class="display">
```
bdrewery: This file is generated, please modify src/bin/poudriere.8 instead

PR #422, opened 2016-11-04
Fixing the relpath() function, not considering partial comon name

pizzamig:
real repro case:
relpath /appdata/poudriere/data/.m/101x64-adm5-default/ref/var/db/ports /appdata/poudriere-etc/poudriere.d/101x64-adm5-options

The output is:
/appdata/poudriere /data/.m/101x64-adm5-default/ref/var/db/ports -etc/poudriere.d/101x64-adm5-options

The proposed patch tries to fix the bug considering a string starting with '/' as good enough


bdrewery:
I pushed some tests out and this change breaks some existing ones.

See: `make -C test`

```
relpath.sh: (common) dir1: '/prefix/a/b/c' dir2: '/prefix/a/b': expected: '/prefix/a/b', actual: '/prefix/a'
```

bdrewery:
Fixed, thanks!

PR #426, opened 2016-11-29
Honoring the -p option in the options command and adding port name based customizations

pizzamig:
This pull request enables several customization based on the port name, fixing the ignored -p option in the option command.

This pull request add support to:
<tree>-<stename>-options
<tree>-<stename>-make.conf
<tree>-<stename>-blacklist
<tree>-<stename>-poudriere.conf

Moreover, if fixes a small bug using <name>-<setname>-blacklist customization.

"poudriere options" still have a problem: it can be invoked without -j option (-p or -z of both); in this case, no architecture is specified, because normally is obtained by the jail.
If no jail is defined, I'm using the host architecture, but a specific new option can be implemented to specify an architecture when no jailname is provided.


bdrewery:
Thanks for this. I will review it soon.

pizzamig:
The biggest open point is using the options command without the jailname: the jail is used to identify the architecture and some options are architecture dependent.
I'd like to add an option, mutually exclusive to the jailname, to specify the architecture. What do you think?

src/share/poudriere/common.sh
```diff
@@ -1861,8 +1863,10 @@ load_blacklist() {
 	local setname=$3
 	local bl b bfile
 
-	bl="- ${setname} ${ptname} ${name} ${name}-${ptname}"
-	[ -n "${setname}" ] && bl="${bl} ${bl}-${setname} \
+	bl="- ${setname} ${ptname} ${name}"
+	[ -n "${setname}" ] && bl="${bl} ${ptname}-${setname}"
+	bl="${bl} ${name}-${ptname}"
+	[ -n "${setname}" ] && bl="${bl} ${name}-${setname} \
```
bdrewery: Nice, this also fixes broken jail-set blacklist support.

bdrewery: Add I fixed this in `release-3.1` in 63b6b0618 but forgot to bring it into `master`.

bdrewery:
For the architecture, rather than assume to use host arch I would prefer adding an `-a arch` option like `jail` has.

```
a)
        ARCH=${OPTARG}
        # If TARGET=TARGET_ARCH trim it away and just use
        # TARGET_ARCH
        [ "${ARCH%.*}" = "${ARCH#*.}" ] && ARCH="${ARCH#*.}"
        ;;
```

pizzamig:
then I add the architecture option support!

pizzamig:
Added a `-a arch` option to the `options` command to handle different architecture like `jail`
If a `-j` and `-a` are used, a consistency check is performed
`setup_makeconf` defines an architecture accordingly

src/bin/poudriere.8
```diff
@@ -735,6 +735,12 @@ A list of ports can be specified directly.
 .Pp
 This command accepts the following options:
 .Bl -tag -width "-f conffile"
+.It Fl a Ar architecture
+Specifies which
+.Ar architecture
+of FreeBSD to use to show options. (Default: the one of jail
```
bdrewery: Need newline after period

src/share/poudriere/common.sh
```diff
@@ -2320,11 +2324,15 @@ setup_makeconf() {
...
+	get_host_arch host_arch
 	# The jail may be empty for poudriere-options.
 	if [ -n "${name}" ]; then
 		_jget arch "${name}" arch
-		get_host_arch host_arch
+	elif [ -n "$ARCH" ]; then
+		arch=$ARCH
+	fi
 
+	if [ -n "$arch" ];
```
bdrewery: Missing `then` after `;`

pizzamig:
Branch merged with master, fixed the errors and performed some tests, using security/openssl as template and i386 and amd64 as given architecture

bdrewery:
ok to test

PR #427, opened 2016-12-17
Allow stage-qa to fail if there are proxydep errors

derekschrock:
ports Mk/Scripts/qa.sh will not return 1 if there's proxy dependencies
unless PROXYDEPS_FATAL is set.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -2411,7 +2411,7 @@ _real_build_port() {
 			local die=0
 
 			bset_job_status "stage-qa" "${port}"
-			if ! injail env DEVELOPER=1 ${PORT_FLAGS} \
+			if ! injail env PROXYDEPS_FATAL=1 DEVELOPER=1 ${PORT_FLAGS} \
 			    /usr/bin/make -C ${portdir} stage-qa; then
 				msg "Error: stage-qa failures detected"
 				[ "${PORTTESTING_FATAL}" != "no" ] &&


derekschrock:
Snippet from a testport with failing proxydeps:

```
....
===========================================================================
====> Running Q/A tests (stage-qa)
Error: /usr/local/bin/mutt is linked to /usr/local/lib/libassuan.so.0 from security/libassuan but it is not declared as a dependency
Warning: you need LIB_DEPENDS+=libassuan.so:security/libassuan
Error: /usr/local/bin/mutt is linked to /usr/local/lib/libgpg-error.so.0 from security/libgpg-error but it is not declared as a dependency
Warning: you need LIB_DEPENDS+=libgpg-error.so:security/libgpg-error
*** Error code 1

Stop.
make: stopped in /usr/ports/mail/mutt
====>> Error: stage-qa failures detected
build of mail/mutt ended at Fri Dec 16 20:19:44 EST 2016
build time: 00:00:23
!!! build failure encountered !!!
```

derekschrock:
After updating the port with the proper LIB_DEPENDS:

```
===========================================================================
====> Running Q/A tests (stage-qa)
====> Checking for pkg-plist issues (check-plist)
===> Parsing plist
...
```

mat813:
No, no, the idea is that if you want the proxydep qa check to error out, you put:

PROXYDEPS_FATAL=true in poudriere's make.conf.

Right now, it is too early to make it really fatal, way too much ports will fail because of it.

derekschrock:
Fair enough.  I was going to comment on the fall out of this but you beat me to it.

PR #429, opened 2016-12-20
Allow using the base system as a jail

mat813:
# (work in progress)

So, looking at what synth does, I need to be able to null_mount, or clone with cpdup, or something, those:

```
tmpfs                   /some/where
/bin                    /some/where/bin
/boot                   /some/where/boot
tmpfs                   /some/where/boot/modules
devfs                   /some/where/dev
/lib                    /some/where/lib
/libexec                /some/where/libexec
/sbin                   /some/where/sbin
/usr/bin                /some/where/usr/bin
/usr/include            /some/where/usr/include
/usr/lib                /some/where/usr/lib
/usr/lib32              /some/where/usr/lib32
/usr/libdata            /some/where/usr/libdata
/usr/libexec            /some/where/usr/libexec
/usr/ports              /some/where/usr/ports
/usr/ports/distfiles    /some/where/distfiles
tmpfs                   /some/where/usr/local
/usr/sbin               /some/where/usr/sbin
/usr/share              /some/where/usr/share
/usr/src                /some/where/usr/src
/var/db/options         /some/where/options
/usr/ports/packages     /some/where/packages
```

This needs to happen in a special case of the "else" of clonefs, but I am unsure of how to proceed.


bdrewery:
Thanks for this, I'll review soon.

Duffyx:
Hello, any news/progress on this?

bdrewery:
Yes it is definitely wanted but no longer applies after refactoring.

Duffyx:
Would it help if I resubmit the patches with another PR against the current tree?

bdrewery:
It might help but it was not 100% complete. Poudriere makes assumptions that it can *rollback* anything it does during the build which won't hold true for this method. So a lot of care is needed to ensure the */* system isn't trashed. And Poudriere assumes there's no packages installed in the "jail" when it starts building, so there's likely going to be several issues with actually building ports.

Duffyx:
I see, the level of complexity is far above my head ... I won't be able to help with this.
Would be an awesome feature though... but I know you guys are really busy!

PR #432, opened 2017-01-22
QEMU: Use host pkg-static

MikaelUrankar:
When building for armv6 the 'package phase' uses the armv6 binary [1] and the build can fail for large package (Fails to install: runaway process)

We can use the host pkg-static to avoid using qemu.

[1] ps ax:
/usr/local/bin/qemu-arm-static /usr/local/sbin/pkg-static create -r
/wrkdirs/usr/ports/print/texlive-texmf/work/stage

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -1795,6 +1795,10 @@ jail_start() {
 		msg "Copying latest version of the emulator from: ${EMULATOR}"
 		mkdir -p "${tomnt}${EMULATOR%/*}"
 		cp -f "${EMULATOR}" "${tomnt}${EMULATOR}"
+		# Copy in the latest version of pkg-static
+		msg "Copying latest version of pkg-static from: ${LOCALBASE:-/usr/local}/sbin/pkg-static"
+		cp -f "${LOCALBASE:-/usr/local}/sbin/pkg-static" "${tomnt}/usr/sbin/pkg-static"
+		echo "PKG_BIN=/usr/sbin/pkg-static" >> "${tomnt}/etc/make.conf" 
 	fi
 	# Handle special ARM64 needs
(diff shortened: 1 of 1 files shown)


bdrewery:
This is already done.

The /.p/pkg-static that is ran in the jail comes from the host:

```
ensure_pkg_installed() {
        local force="$1"
        local mnt

        _my_path mnt
        [ ${PKGNG} -eq 1 ] || return 0
        [ -z "${force}" ] && [ -x "${mnt}${PKG_BIN}" ] && return 0
        # Hack, speed up QEMU usage on pkg-repo.
        if [ ${QEMU_EMULATING} -eq 1 ] && \
            [ -f /usr/local/sbin/pkg-static ]; then
                cp -f /usr/local/sbin/pkg-static "${mnt}/.p/pkg-static"
                return 0
        fi
        [ -e ${MASTERMNT}/packages/Latest/pkg.txz ] || return 1 #pkg missing
        injail tar xf /packages/Latest/pkg.txz -C / \
                -s ",/.*/,.p/,g" "*/pkg-static"
        return 0
}
```

bdrewery:
At least for pkg-repo, for the "package phase" it was tested and did not work for all ports reliably, or was creating bad packages. The code for that is present but disabled in `_real_build_port`:

```
# Use bootstrap PKG when not building pkg itself.      
if false && [ ${QEMU_EMULATING} -eq 1 ]; then          
        case "${port}" in                              
        ports-mgmt/pkg|ports-mgmt/pkg-devel) ;;        
        *)                                             
                if ensure_pkg_installed; then          
                        export PKG_BIN="/.p/pkg-static"
                fi                                     
                ;;                                     
        esac                                           
fi                                                     
```

bdrewery:
I forget what the problem was exactly, but we do have the code for it. Retesting it on a build, and the resulting packages would be helpful. From there we could do an exp-run to see if it works for the whole tree. Just remove the `false &&`.

PR #433, opened 2017-01-24
switch to http for ftp-archive

lifanov:
This is to work around FTP protocol errors fetching some old releases.


bdrewery:
It seems simpler to just change ftp:// to http:// for the archive.

lifanov:
Sure! I was just taking a "first, do no harm" approach.

lifanov:
Which one would you prefer: global switch to http:// or a new method?

bdrewery:
On 2/1/2017 10:50 AM, Nikolai Lifanov wrote:
> Which one would you prefer: global switch to http:// or a new method?
> 

If http:// works fine then I think we should just switch to that. It's
more likely to work in various firewalls/environments than ftp://.

bdrewery:
Can you squash these commits together please? No sense having a revert in the history if not needed. It makes git-blame more difficult.

PR #435, opened 2017-02-08
invoke TEST_TARGET for testport

iblislin:
ref: https://svnweb.freebsd.org/ports/head/Mk/bsd.port.mk?revision=433456&view=markup#l842


bdrewery:
This is not enough as TEST_DEPENDS needs to be installed. This is a duplicate of #355.

PR #436, opened 2017-02-08
do not copy src directory if -S option is also given when src method is used

avg-I:
This is useful for a scenario where I want to create a jail from a pre-built source tree, but for the builds that will use the jail I want to provide a different source tree.

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -416,22 +416,24 @@ build_and_install_world() {
 install_from_src() {
 	local cpignore_flag cpignore
 
-	msg_n "Copying ${SRC_BASE} to ${JAILMNT}/usr/src..."
-	mkdir -p ${JAILMNT}/usr/src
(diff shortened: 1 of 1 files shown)


bdrewery:
The patch looks fine, but why would you want a different source tree in the jail than what it was installed from?

avg-I:
As I understand, /usr/src is needed mostly for kernel modules and tools that talk to kernel. I use FreeBSD head, I update the kernel quite often, but the world not that often (because of how long it takes to build it).
So, I want my poudriere to pick the main (in sync with the running kernel) /usr/src via nullfs during builds.  But when I rebuild the world I want to be able to use -m src=/usr/src to install or update the jail.

bdrewery:
Are you using these jails for something other than port builds?

For port module builds the /usr/src needs to match the target kernel (and be stable for rebuilds where the jail was not updated).  I suppose if you were using this jail for port builds to use packages locally you would always want the latest host /usr/src mounted in for kernel packages.

PR #449, opened 2017-02-19
Adapt queue command to send the correct arguments to the poudriered

madpilot78:
socket.

### DIFF
--- a/src/share/poudriere/queue.sh
+++ b/src/share/poudriere/queue.sh
@@ -26,21 +26,18 @@
 # SUCH DAMAGE.
 
 usage() {
-	echo "poudriere queue queuename poudriere_command"
+	echo "poudriere queue poudriere_command"
 	exit 1
 }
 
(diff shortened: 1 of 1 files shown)


src/share/poudriere/queue.sh
```diff
@@ -26,21 +26,18 @@
...
 
 case $1 in
-bulk|testport) ;;
+bulk|testport) cmd=$1 ; shift ;;
 *) err 1 "$2 command cannot be queued" ;;
 esac
 
 # Queue the command through the poudriered socket
-echo "${name} POUDRIERE_ARGS: $@" | nc -U ${QUEUE_SOCKET}
+echo "opeeration = queue; command = $cmd; arguments = $@;" | nc -U ${QUEUE_SOCKET}
```
bdrewery: Oops 'operation' spelled wrong. It's ok though I am fixing it.

PR #453, opened 2017-03-12
Introduce switch to select if HTML files are inline or hosted

gbeine:
It is sometimes quite annoying if the HTML files are sed'ed with "inline".

This switch allows the user to select in poduriere.conf if this should happen.


bdrewery:
Nice, thanks!

Issue #441, opened 2017-02-15
Make poudriere jail -x automatic


bdrewery:
This can be determined automatically as it is in the FreeBSD.org package build script:
```
needs_qemu=$(/sbin/sysctl -n kern.supported_archs | grep -qw "${arch#*.}" || \
    echo yes)
if [ "${needs_qemu}" = "yes" ]; then
        native_xtools="-x"
else
        native_xtools=
fi
```

bdrewery:
Even if this is not done, a warning should be given to the user about not having native-xtools installed in `bulk`.

Issue #445, opened 2017-02-16
XDEV hardlinks should be done at jail startup time


bdrewery:
It's not ideal that if we fix a bug here that users need to recreate their jails. Just do this stuff at jail startup time.

bdrewery:
Need #447 first

PR #455, opened 2017-03-22
(WIP) Default to building native-xtools

kevans91:
Creating this mostly for review purposes. native-xtools building through the -x option offers a fairly nice performance benefit with no downside that I've seen. It's also been  recommended to every instance I've seen of queries involving cross-build poudriere jails. 

I've based this off of issue #441, but (ab)using the fact that QEMU_EMULATING is already set by check_for_emulation in every path I examined. I've removed the documentation for -x since it will be the default, and otherwise note the following:

1.) I've separated this out into a check_xdev to do this and run some basic sanity checks to catch basic reasons for failure of building xtools. This seemed like a good idea, given it's the default and it sucks to get all the way through the build/install process just to find that that part failed.

2.) I call check_xdev in build_and_install_world. I think this might be a little too late to call this, though? I can't immediately think of a reason /usr/src may not be a working src tree before -m is handled, but I put it afterwards to play it safe until I could consult someone more familiar. If we can check earlier, that would be more ideal than checking out/copying sources when it's a guaranteed fail.

3.) Would it make sense to make src tree on host system (currently /usr/src) configurable globally or per-jail? I'm afraid I'm not familiar enough with native-xtools call to determine this, but it seems like that would a good thing to add. On my two build systems, I personally maintain /usr/src for -head and /usr/src.11 for stable/11. These are generally used for pkgbase builds, but I also run both stable/11 and -head poudriere builds.

It may not be the case at the moment, but it is possible that stable/11 and -head native-xtools targets could produce different enough results that it would be more desirable to specify that my stable/11 jails instead use /usr/src.11 for native-xtools, right?

If so, I can look into making this configurable. It could fortunately default to /usr/src and switch to PATH if -m src=PATH, so hopefully most people wouldn't need to touch it.


bdrewery:
I think I like this. It will need to be rebased after my work for #445 and #477 comes in.  The `check_xdev` call just needs to move to the top of a new `build_native_xtools` function in _jail.sh_.

bdrewery:
ok to test

bdrewery:
This can be rebased now.

kevans91:
That simplified a lot, should be pretty easy to review now. =)

bdrewery:
Thanks!

bdrewery:
My current thoughts are that I want to fix `make native-xtools` having to run from /usr/src before forcing this on.

However, I'm now thinking that "fixing" native-xtools to use the jail source would actually be wrong.  We want binaries that the host kernel can run, not newer ones with newer syscalls.  Though that same problem already exists for all Poudriere usage, that you have to run a jail the same or older than the host. So forget that thought.

I do think we should have `make native-xtools` fixed first so having something like this forced one "just works" rather than throwing up surprises about _/usr/src_ being needed.  I have some WIP items for that.

PR #456, opened 2017-03-22
do not install package in interactive mode 

capjo:
Specified packages are installed after entering the 'phase: Interactive' by default.

Some packages, e. g. tex, are big and take time to install. To avoid this overhead it would be helpful to to be able to control this behavior.

This minimalistic patch uses the environment variable "POUDRIERE_INTERACTIVE_NO_INSTALL" to specify that the specified packages should not be installed when entering the interactive phase.


bdrewery:
This still installs dependencies, just not the main package.  Is that what you wanted?

capjo:
Yes it is. It would be great to have it.

mat813:
This makes me wonder, couldn't poudriere save up on build time by not installing the RUN_DEPENDS ?

I mean, RUN_DEPENDS are not needed to build the package (which is the goal of poudriere) they are only needed to run the packages after being installed.  The only thing that is needed is to register the dependency.

capjo:
Having the runtime dependencies to check the build and the compiled binaries would be useful to me, but the installation is simply not necessary.

phase deinstall: only removes the packages specified to build, but keeps the runtime dependencies installed.
phase interactive: installs the specified packages again (not necessary)

bdrewery:
What is the point of installing dependencies but not the main package?

AndiDog:
Installing dependencies but not the main package gives you the ability to manually test with executables/libraries/outputs that were built in WRKDIR and therefore saves the time to extract the package to /usr/local. I think that can become _one purpose_ of the interactive mode.

capjo:
The dependencies are necessary to run tests that are part of the build process.

Another use case is: modify the package sources -> compile changes -> run tests

If the package is already installed you wouldn't know which shared objects, or include files, the tests will use without looking into the details (build directories vs. system directories).

grembo:
@bdrewery: As I'm involved in the discussion offline as well: The idea is a workflow in which the port is built using _poudriere_ in advanced interactive mode and keep the long running jail open for further port development server side. So the developer pokes into jailname-n and modifies the port skeleton and/or ports sources and builds the port again within the jail, runs various test etc. until development is finished and the build jail is stopped.

This is done using multiple package sets, so that builders can run independently to work on different ports in parallel over time (so e.g. you would have 103amd64-latex-latex, 103amd64-chrome-chrome running in parallel while working on ports trees and ports in parallel). As having the package already installed might have side-effects (like @capjo described), it's preferred to have all the dependencies in there, but not the package in question. Also running `pkg delete` every time an initial build is done can be forgotten easily and for big ports it's adding a time consuming add/delete cycle for no reason.

It might make sense to make this feature more fine grained though, so one could also skip installation of RUN_DEPENDS like you suggested, I could see some use in that (even though it wouldn't help the particular use case I'm talking about).

bdrewery:
Thanks, I'll take it when I find some time to focus on Poudriere again in the next few weeks.

bdrewery:
Ok I see the use for this now. I'm not fond of more options though. Not sure how else this would be addressed though.

Is this all actually interactive? We could have a confirmation prompt ask when entering the interactive jail whether dependencies and the package should be installed.

AndiDog:
> Is this all actually interactive? We could have a confirmation prompt ask when entering the interactive jail whether dependencies and the package should be installed.

In our use case it's _not_ interactive, so there should be some option (env var, command line arg, ...) to control the behavior. Essentially, we run a full poudriere build once, tell it to leave the build jail running (`poudriere testport -I`) and then use the build jail via scripts (e.g. build with changed source code). This follows existing behavior of `INTERACTIVE_MODE == 2` in that it does _not_ ask any questions or enter an interactive prompt.

PR #460, opened 2017-04-12
Do not deadlock with random PIDs.

dankm:
Fix a race condition where pwait would hang if passed its own pid. This
could happen if kern.randompid is set and a the intended waitee pid gets
recycled before pwait got called.

Instead of deadlocking just exit immediately.

This mirrors the bug I filed with FreeBSD at https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=218598


bdrewery:
Thanks. Let's see what happens with the FreeBSD base version first as I try to keep in-sync with it as much as possible. This is a bit of a silly case that I suspect only Poudriere would be hitting.

PR #466, opened 2017-04-26
jail: Add -D option to keep git history

evadot:
While it's useful to clone without history by default, sometime you
need the full one. For example if you rely on git tags to generate an image.

Signed-off-by: Emmanuel Vadot <emmanuel.vadot@gandi.net>


bapt:
Do not forget the manpage :)

PR #469, opened 2017-05-17
add option to delete portstree when created with -F

danel1:
When creating a portstree with the following options:
"poudriere ports -c -F -p name"
everything get's created correctly and the zfs is also made.
But since the following two commit's, it's inpossible to delete the portstree with poudriere. The only way would be to delete and clean up everything by hand, which is not a nice way and the way you want to do it.

See:
https://github.com/freebsd/poudriere/commit/ba02294a1f4aacc3f54e544178c92628c8e4b7f6
and
https://github.com/freebsd/poudriere/commit/7ca6e58de9958eb70c72825a62438d9ed8d3bead

### DIFF
--- a/src/share/poudriere/ports.sh
+++ b/src/share/poudriere/ports.sh
@@ -44,6 +44,8 @@ Options:
                      to 'head/master'.
     -F            -- When used with -c, only create the needed filesystems
                      (for ZFS) and directories, but do not populate them.
+    -x            -- Force delete portstree if created with "-F" and
+                     method is ="none" or ="-". Can only be used with "-d".
     -M path       -- The path to the source of a ports tree.
     -f filesystem -- The name of the filesystem to create for the ports tree.
                      If 'none' then do not create the filesystem.  The default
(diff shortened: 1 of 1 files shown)


bdrewery:
Thanks for the patch. I'm not fond of another flag here. The root problem is that `-F` does something and never records the fact that it did it. If `-F` was used (without method `-` or `none`) then it needs to automatically clean up.

There's far too many flags and options here. `-m none` and `-f none` do totally different things, and `-F` is a hack.

Looking over this more, it seems that using `-F` will actually record the method as `-`, so 7ca6e58de9958eb70c72825a62438d9ed8d3bead may just need to be reverted instead.

bdrewery:
For reference, further discussion at ba02294a1f4aacc3f54e544178c92628c8e4b7f6.

danel1:
True, i just tought this would also be a fast way to patch this, instead of completely rewrite the ports.sh-Script.. :-) But indeed, if a portstree is created with "-F", it always uses method "-". So reverting 7ca6e58 would be the most simple and "best" option until we find a better solution, i think.

bdrewery:
I missed saying it, but I am tempted to rip out all of this obscure stuff and simply the interface. And then add stubs to disallow the old ones.

danel1:
Thanks for adding the milestone @bdrewery! I hope you can fix this in 3.1.20. Have a good day

bdrewery:
Yup I am going to take a fresh look at it in the next week.

bdrewery:
This is fixed for newly-created trees after these fixes.

danel1:
Thanks :) Will take a look at it and test when the new version will be released.
Do you have a timeline when you plan to release 3.1.20?

Issue #477, opened 2017-06-04
jail -x needs to be supported for non-src builds


bdrewery:
`jail -c -j 120C-SNAP-aarch64 -x -m url=http://ftp.freebsd.org/pub/FreeBSD/snapshots/arm64/aarch64/12.0-CURRENT/ -v 12.0-CURRENT -a arm64.aarch64`

This did not actually build and install the XDEV / native-xtools pieces.

rootwyrm:
Yep... XDEV=1 only triggers within `build_and_install_world()` at line 364 in jail.sh. What would the preference be as far as addressing that? Should only need 364-416. But given various factors, would it make sense to be able to rebuild xtools outside of jail creation (e.g. as part of update)?

bdrewery:
Update usually calls `build_and_install_world` too. We need the XDEV code to run anytime `-c` or `-u` that `-x` (`XDEV==1`) is set, before the `@clean` snapshot is created.

bdrewery:
I would not bother doing it for `-m null` though and instead have an error that it is not supported.  It's best to not mess with an existing chroot and to consider it read-only as much as possible.

rootwyrm:
As implemented, users performing updates will need to include `-X`, but it should cover clean and update scenarios.

PR #478, opened 2017-06-04
Add support for -X to build xtools without buildworld

rootwyrm:
#477 

Improves poudriere to build xtools with -X for all methods other than SRC and VCS

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -69,6 +69,10 @@ Options:
                      building for a different TARGET ARCH than the host.
                      Only applies if TARGET_ARCH and HOST_ARCH are different.
                      Will only be used if -m is svn*.
+    -X            -- Build and setup native-xtools cross compile tools in jails
+                     which were built from ISO or snapshot for a different TARGET
+                     arch than HOST_ARCH.
+
 
 Options for -s and -k:
     -p tree       -- Specify which ports tree to start/stop the jail with.
@@ -273,6 +277,11 @@ update_jail() {
 		err 1 "Unsupported method"
 		;;
 	esac
+
+	if [ ${XDEV} -eq 1 ]; then
+		install_xdev_standalone
+	fi
+
 	jset ${JAILNAME} timestamp $(clock -epoch)
 }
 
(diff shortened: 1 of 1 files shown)


src/share/poudriere/jail.sh
```diff
@@ -1008,6 +1092,9 @@ while getopts "biJ:j:v:a:z:m:nf:M:sdkK:lqcip:r:uU:t:z:P:S:Dx" FLAG; do
 		x)
 			XDEV=1
 			;;
+		X)
+			XDEV=1
```
bdrewery: Er, -X does the same thing as -x in your patch. I don't see the need for a new flag anyway. -x should be sufficient for all cases. Just revert the getops/X)/usage changes.

src/share/poudriere/jail.sh
```diff
@@ -531,7 +606,7 @@ install_from_vcs() {
 		git_sha=$(git -C ${SRC_BASE} rev-parse --short HEAD)
 		version_vcs="${git_sha}"
 	;;
-	esac
+	esac
```
bdrewery: Whitespace change

src/share/poudriere/jail.sh
```diff
@@ -273,6 +277,11 @@ update_jail() {
 		err 1 "Unsupported method"
 		;;
 	esac
+
+	if [ ${XDEV} -eq 1 ]; then
```
bdrewery: This comes too late, note line 261 where the jail is marked clean.  It's the `@clean` zfs snapshot which is rolled back to at startup.  You'll need to hook `install_xdev_standalone` from where it used to be called.

bdrewery: Call it near line 243 as well for ftp/snapshot builds.

src/share/poudriere/jail.sh
```diff
@@ -436,6 +445,72 @@ build_and_install_world() {
 	fi
 }
 
+install_xdev_standalone() {
```
bdrewery: It looks like you copied this from the existing place - the existing place should be removed and just call this.

bdrewery:
I'm covering this with a big patch. Thanks though.

Issue #415, opened 2016-09-23
JSON/HTML tracks by origin but should be pkgname


bdrewery:
Need to switch to pkgname to resolve #259

Issue #313, opened 2015-05-18
Poudriere fails to build pyNN- dependencies of non-default python version


yurivict:
Port shar is here: http://pastebin.com/c98Jn1XX

It does rebuild and extract python34 fine:

```
[10amd64-local-workstation-job-01] `-- Extracting libffi-3.2.1... done
[10amd64-local-workstation-job-01] Extracting python34-3.4.3_1... done
===>   Returning to build of tiny-network-utilities-0.150518
```

but it fails to rebuild required py34-\* modules:

```
=======================<phase: run-depends    >============================
===>   tiny-network-utilities-0.150518 depends on package: py34-netifaces>0 - not found
===>    Verifying install for py34-netifaces>0 in /usr/ports/net/py-netifaces
===>   tiny-network-utilities-0.150518 depends on package: /packages/All/py34-netifaces-0.10.3.txz - not found
===>   USE_PACKAGE_DEPENDS_ONLY set - not building missing dependency from source
*** Error code 1
```

It looks like poudriere rebuilds py27- dependencies, and then fails on missing py34- ones.

This may well be the problem in ports Mk/, but it only fails in poudriere.

yurivict:
Looking into this myself. Looks like pourdiere isn't package generation aware.

'list_deps' is where this begins. Changing it to this:
````
list_deps() {
        [ $# -ne 1 ] && eargs list_deps directory
        local dir="/usr/ports/$1"
        local makeargs="-VPKG_DEPENDS -VBUILD_DEPENDS -VEXTRACT_DEPENDS -VLIB_DEPENDS -VPATCH_DEPENDS -VFETCH_DEPENDS -VRUN_DEPENDS"

        prefix_stderr_quick "(${COLOR_PORT}$1${COLOR_RESET})${COLOR_WARN}" \
                injail make -C ${dir} $makeargs | \
                tr ' ' '\n' | \
                sed -E "s,^(py[0-9]+)-(.*)/py-(.*),\1-\2/\1-\3,g" | \
                sed -e "s,[[:graph:]]*/usr/ports/,,g" \
                -e "s,:[[:graph:]]*,,g" -e '/^$/d' | \
                sort -u || err 1 "Makefile broken: $1"
}
````
makes it compute package names with generations.
This will need to be used to strip generation:
````
pname_strip_generation() {
        echo "$1" | sed -E "s,^(py[0-9]+)-,py-,g" | \
}
````
[...]

PR #483, opened 2017-06-06
Support DEPENDS_ARGS/FLAVORS - Building multiple packages from one origin

### LINKED ISSUE #259: Provide support for passing build flags down the dependency chain

bdrewery:
This fixes #259 and #313, and obsoletes #314.

Remaining items:
1. ~The next thing will be to add _FLAVORS_ support for https://reviews.freebsd.org/D10327.  This should mostly be trivial.~
2. ~Support ORIGIN@FLAVOR syntax for `bulk -f list` and `bulk port`, etc via `listed_ports`
   Ensure that ORIGIN ORIGIN@FLAVOR1 ORIGIN@FLAVOR2 works together.~
3. ~Support _RUN_DEPENDS+= foo>0:devel/foo:FLAVOR=flave_~
4. ~Make `is_bad_flavor_slave_port` requeue the main port with the proper `DEPENDS_ARGS=PYTHON_VERSION` for the originally request port.  This will relieve the pressure to commit _FLAVORS_ support immediately to prevent breaking all of the python ports listed.~
5. ~The chosen FLAVOR or DEPENDS_ARGS is not always clear in the web interface but should be computable from the log. Not sure I will modify this before merge.~ (Will handle in #415)
6. ~All `cache_get_origin` need auditted to ensure they shouldn't be using originspec, like bulk showing ports to build.~
7. ~The reinjection of items into the queue for DEPENDS_ARGS and FLAVORS needs to be unified for simplicity.~ **moved to #492**
8. ~Similar to (7), _FLAVORS_ and _DEPENDS_ARGS_ should not be supported set together.~
9. ~Support `origin@-` to only build the first default flavor.~
10. ~Need a test suite desperately.  There's so many test cases here that keep failing when I change things. #488~
11. ~Fix the `CHECK_CHANGED_DEPS` support for _ORIGIN@FLAVOR_ dependency lines.  It currently will bomb since the packages only record _ORIGIN_.  Need to convert to _PKGBASE_ or something.  This code belongs in ports badly. #495~

~The python slave port issue is quite complex.
If a port like _net/tiny-network-utilities_ requires _net/py-netifaces_ as python3 it currently does not work due to the _DEPENDS_ARGS_ bug.  The workaround today in ports is to change the dependency to a new slave port such as _net/py3-netifaces_.  In this specific example _net/tiny-network-utilities_ does not depend on a py3 slaveport.  Once Poudriere is merged with _DEPENDS_ARGS_ support it, with `bulk -a` will then queue the virtual _net/py-netifaces_ for python2 (default) and then also python3 from the _net/tiny-network-utilities_ _DEPENDS_ARGS_ dependency.  It will also queue a build for _net/py3-netifaces_.  The slave port and the dependency-generated package will both be 'py3-netifaces' for PKGNAME and thus hit the duplicated PKGNAME assertion and fail to build.
To work around this I force mapped origins that had a slave port added but would build just fine now from the main port from a DEPENDS_ARGS-generated package.  The problem that comes from all of this is that if a py3 slave port was added not for a dependency solution but purely to provide a py3 package for a user.  In that case without FLAVORS a py3- package will never generate from the default port.  So merging this into the package builds will drop a lot of py3- packages.  That is fixed by merging FLAVORS support to the python ports which then will ensure all main python ports provide a py2 and py3 package from the origin whether they are a dependency or not.~

Performance of `bulk -a` startup is hurt by 7% but it's relatively fast still.

```
x before
+ after
+------------------------------------------------------------+
|                                                           +|
|   x                                                       +|
| x x      x                                                +|
||__MA___|                                                   |
|                                                           A|
+------------------------------------------------------------+
    N           Min           Max        Median           Avg        Stddev
x   4        275.69        279.24        276.52      276.9925     1.5485773
+   3           297           297           297           297             0
Difference at 95.0% confidence
        20.0075 +/- 2.35542
        7.22312% +/- 0.911779%
        (Student's t, pooled s = 1.19952)
```


bdrewery:
A lot of spam, I know. I think I'm down to the last bug with `CHECK_CHANGED_DEPS` supporting _ORIGIN@FLAVOR_ lines.

swills:
run tests

swills:
run tests

swills:
run tests

swills:
run tests

swills:
retest this please

swills:
retest this please

swills:
retest this please

swills:
retest this please

swills:
retest this please

swills:
retest this please

swills:
please retest this

swills:
retest this please

bdrewery:
retest

bdrewery:
retest this please

bdrewery:
retest this please

bdrewery:
retest this please

bdrewery:
retest this please

PR #486, opened 2017-06-08
poudriere image: Add embedded type

evadot:
The sdcard type is mostly targeted for arm,arm64 and mips boards.
It create an 20M fat16 partition for storing u-boot, loader and possibly
DTB at one point.
As target arch might be different than the host arch, packages are installed
with pkg -r and specifiying the ABI in the env.
This is a BYOU (Bring You Own U-Boot) image, intalling U-Boot is, for now,
beyond the scope.


bapt:
Why naming that sdcard when it would also fit on a memdisk, or even normal disk?
What about a ubootimage or something like that instead?

evadot:
ubootimage would be confusing as U-Boot as an image format named uImage.
sdcard is the best name I can come up with but I'm ok to change it to something more appropriate.

PR #490, opened 2017-06-10
poudriere image: fix iso generation

evadot:
ISO images needs the boot loaders to be cpdup into a directory that makefs
can use.
makefs need to use the correct directory with world+kernel, not just /boot
And the image need a fstab to the kernel knows that it needs to mount the
cdrom as root device.


evadot:
Do not merge, this break iso+mfs image creation, I'll check why.

evadot:
False alarm, everything is fine.

PR #513, opened 2017-07-08
Expand xdev support and setup hardlinks at jail startup time.

### LINKED ISSUE #445: XDEV hardlinks should be done at jail startup time

### LINKED ISSUE #477: jail -x needs to be supported for non-src builds

bdrewery:
- Expand `jail -x` support to methods: _tar_, _src_ without `-b`, and _ftp_ type methods. Fixes #477 and obsoletes #478 
- Xdev hardlinks are now setup at jail start time rather than `jail -cu` time.  This allows less modifications to the jail's snapshot and not requiring rebuilding a jail if Poudriere decides to modify the _HLINK_FILES_ list.  Fixes #445

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -2089,6 +2089,65 @@ qemu_install() {
 	cp -f "${EMULATOR}" "${mnt}${EMULATOR}"
 }
 
+setup_xdev() {
+	[ $# -eq 2 ] || eargs setup_xdev mnt target
+	local mnt="$1"
+	local target="$2"
+	local HLINK_FILES file
+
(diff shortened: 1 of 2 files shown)


bdrewery:
retest

bdrewery:
retest this please

PR #514, opened 2017-07-09
Fix LOIP4/6 checks which always fail when using poudriere in a jail

mhalden:
The check for if address from LOIP4 and LOIP6 are using the wrong addresses so they will always fail. Because of this no address is assigned to the local only jail and as a result building a lot of packages failed to build, including python and ruby.

This PR fixes the checks so it actually looks for an inet address with LOIP4 and an inet6 address with LOIP6, this was reversed earlier.


swills:
OK to test

bdrewery:
Hmm, I'll review this.

bdrewery:
Thanks!

PR #515, opened 2017-07-24
Show QEMU version in buildlog

MikaelUrankar:
Hi,
For debugging purpose it would be nice to print the qemu version in the log file.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -684,6 +684,7 @@ buildlog_start() {
 	echo "maintained by: $(injail /usr/bin/make -C ${portdir} maintainer)"
 	echo "Makefile ident: $(ident -q ${mnt}/${portdir}/Makefile|sed -n '2,2p')"
 	echo "Poudriere version: ${POUDRIERE_VERSION}"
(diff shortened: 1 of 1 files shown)


src/share/poudriere/common.sh
```diff
@@ -684,6 +684,7 @@ buildlog_start() {
 	echo "maintained by: $(injail /usr/bin/make -C ${portdir} maintainer)"
 	echo "Makefile ident: $(ident -q ${mnt}/${portdir}/Makefile|sed -n '2,2p')"
 	echo "Poudriere version: ${POUDRIERE_VERSION}"
+	[ ${QEMU_EMULATING} -eq 1 ] && echo "QEMU version: `pkg query %v qemu-user-static`"
```
bdrewery: This should be `${PKG_BIN} query ...`. But more importantly we don't know what _qemu-user-static_ is what is being used. It could be any qemu package from any qemu origin.

MikaelUrankar: I'm not sure I understand, we only have emulators/qemu-user-static?

PR #530, opened 2017-09-06
Speed up tar creation by using xz -T 0

swills:


swills:
Not sure if you will like this idea or not, but it definitely speeds up tar creation for me. If you would rather this were optional, I can change it.

reezer:
Do you think piping from tar to xz instead of creating, writing to disk, compressing and then moving would make sense in order to decrease disk IO?

swills:
@reezer Yeah, good point, I probably should do that instead. I'll revise.

bdrewery:
Fine with me, @bapt @brd @evadot any feedback? I don't touch this file.

bapt:
fine to me

evadot:
fine with me too.

brd:
Merged, thanks!

Sorry it took so long, I forgot about this.

PR #430, opened 2016-12-20
jail: Add -C (for clean) option in order to clean poudriere data when deleting a jail

sbz:
This change against branch [release-3.1](/freebsd/poudriere/tree/release-3.1) add a new option -C only used for jail delete (-d) operation in order to clean the poudriere data of the jail when deleting it. If no option -C is used, act as previously.

Options clean -C support different values:
- all
- cache
- logs
- packages
- wkrdis

Usage:

$ poudriere jail -d -j ${jname} -C [all|cache|logs|packages|wrkdirs]

PR #532, opened 2017-09-22
jail: Add -C (for clean) option in order to clean poudriere data when deleting a jailJail clean

sbz:
This change against branch [master](/freebsd/poudriere/tree/master) add a new option -C only used for jail delete (-d) operation in order to clean the poudriere data of the jail when deleting it. If no option -C is used, act as previously.

Options clean -C support different values:
- all
- cache
- logs
- packages
- wkrdis

Usage:

$ poudriere jail -d -j ${jname} -C [all|cache|logs|packages|wrkdirs]

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -141,7 +141,7 @@ delete_jail() {
 	rm -rf ${POUDRIERED}/jails/${JAILNAME} ${cache_dir} \
 		${POUDRIERE_DATA}/.m/${JAILNAME}-* || :
 	echo " done"
-	if [ "${CLEANJAIL}" == "none" ]; then
+	if [ "${CLEANJAIL}" = "none" ]; then
 		return 0
 	fi
 	msg_n "Cleaning ${JAILNAME} data..."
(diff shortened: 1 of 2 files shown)


bdrewery:
ok to test

bdrewery:
#430 was the old PR with older CR discussion.

src/share/poudriere/jail.sh
```diff
@@ -70,6 +70,10 @@ Options:
                      Only applies if TARGET_ARCH and HOST_ARCH are different.
                      Will only be used if -m is svn*.
 
+Options for -d:
+    -C clean      -- Clean remaining data existing in pourdiere data folder.
+                     See poudriere(8) for more details. Can be one of:
+                       all, cache, logs, packages, wrkdirs
```
bdrewery: Something seems off with whitespace here, can you ensure there's no tabs and only spaces for the usage area?

src/share/poudriere/jail.sh
```diff
@@ -136,6 +141,22 @@ delete_jail() {
 	rm -rf ${POUDRIERED}/jails/${JAILNAME} ${cache_dir} \
 		${POUDRIERE_DATA}/.m/${JAILNAME}-* || :
 	echo " done"
+	if [ "${CLEANJAIL}" == "none" ]; then
```
bdrewery: `=`

PR #535, opened 2017-10-02
Poudriere is not able to stage some ports if "security.bsd.hardlink_check_*" != 0

egypcio:
Staging and packaging will fail if 'security.bsd.hardlink_check_uid' or 'security.bsd.hardlink_check_gid' are not set to '0'.

Hope that is the right place to put this kind of check.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -2227,7 +2227,9 @@ jail_start() {
 		err 1 "The ports name cannot contain a period (.). See jail(8)"
 	[ "${setname#*.*}" = "${setname}" ] ||
 		err 1 "The set name cannot contain a period (.). See jail(8)"
-
+	if [ ! $(sysctl -n security.bsd.hardlink_check_uid) -eq 0 -o ! $(sysctl -n security.bsd.hardlink_check_uid) -eq 0 ]; then
+		err 1 "Poudriere is not able to stage some ports if 'security.bsd.hardlink_check_uid' or 'security.bsd.hardlink_check_uid' are not set to '0'."
+	fi
 	if [ -z "${NOLINUX}" ]; then
 		if [ "${arch}" = "i386" -o "${arch}" = "amd64" ]; then
 			needfs="${needfs} linprocfs"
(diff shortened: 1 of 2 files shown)


bdrewery:
This is probably only needed if `BUILD_AS_NON_ROOT` is _yes_?

egypcio:
@bdrewery ,
you were right. nice catch. I just pushed a new version of the file.

Thanks!

src/share/poudriere/common.sh
```diff
@@ -2227,7 +2227,17 @@ jail_start() {
 		err 1 "The ports name cannot contain a period (.). See jail(8)"
 	[ "${setname#*.*}" = "${setname}" ] ||
 		err 1 "The set name cannot contain a period (.). See jail(8)"
-
+	if [ ! $(sysctl -n security.bsd.hardlink_check_uid) -eq 0 -o ! $(sysctl -n security.bsd.hardlink_check_gid) -eq 0 ]; then
```
bdrewery: This works but has 2 problems.
1. It runs `sysctl` twice for every jail startup
2. It cares about these on non-FreeBSD platforms that may not have this feature (DragonFly even though the support for it is nearly dead).

I suggest instead putting something like this into _src/share/poudriere/include/common.sh.freebsd_:
```
HARDLINK_CHECK="$(sysctl -n security.bsd.hardlink_check_uid)$(sysctl -n security.bsd.hardlink_check_gid)"
```
You can't check _BUILD_AS_NON_ROOT` here as it has no default yet.
Then in jail_start change the `if` line to be:
```
if [ -n "${HARDLINK_CHECK}" -a ! "${HARDLINK_CHECK}" = "00" ]; then
...
```

egypcio: > I suggest instead putting something like this into `src/share/poudriere/include/common.sh.freebsd`

Done. Thanks! It really looks more elegant now.

> You can't check _BUILD_AS_NON_ROOT_ here as it has no default yet.

Actually it has. I did test it before pushing the new revision. _BUILD_AS_NON_ROOT_ was configured as '**yes**' - without any change on `poudriere.conf` (its default).

To get its value I added a silly "_echo ${BUILD_AS_NON_ROOT}_" one line before the "_case_" on [src/share/poudriere/common.sh](https://github.com/egypcio/poudriere/blob/083c0ef6ffbdfd428852ea5d93fd3177c76ff3d2/src/share/poudriere/common.sh#L2231). I also had `security.bsd.hardlink_check_gid=1`.

I than changed it to '**yEs**' and checked again (and again, changing it to '**NO**').

bdrewery:
Perfect, thanks!

PR #537, opened 2017-10-13
Add the processed package's origin to the compute_deps_pkg error message.

mat813:
This now says something like this:

```
[00:03:20] Error: compute_deps_pkg failed to lookup pkgname for textproc/py-genshi@python3 processing package py36-rst2html5-1.9.3 from textproc/py-rst2html5@python3 -- Is SUBDIR+=py-genshi@python3 missing in textproc/Makefile and does the port provide the '' FLAVOR?
```

It is helpful because some of the time, the error is not in the dependency, but in the source one.


swills:
OK to test

PR #540, opened 2017-11-06
Image abi fix

swills:
Fix an issue with ALTABI getting set to the host ALTABI. While here, fix the pkg installs getting logged to the host syslog.

### DIFF
--- a/src/share/poudriere/image.sh
+++ b/src/share/poudriere/image.sh
@@ -279,9 +279,8 @@ if [ -n "${PACKAGELIST}" ]; then
 	FreeBSD: { enabled: false }
 	local: { url: file:///${WRKDIR}/world/tmp/packages }
 	EOF
-		abi=$(REPOS_DIR=/${WRKDIR}/world/tmp/ pkg -r ${WRKDIR}/world/ query --file tmp/packages/Latest/pkg.txz '%q')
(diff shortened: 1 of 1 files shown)


swills:
retest this please

swills:
Hmm, was, I think, but isn't now. But if there's a better file to use, please let me know.

Without this, building an i386 image on an amd64 host fails, at least for me.

evadot:
It's not anymore (tested on jail with and without native-xtools to be sure)

bdrewery:
Actually only on mips is /bin/sh not a hardlink to /nxb-bin/bin/sh.  All other platforms get the hardlink. Also the hardlinks are only created at *jail startup* now, so just looking in the jail directory without running isn't enough.

```
# Endian issues on mips/mips64 are not handling exec of 64bit shells
# from emulated environments correctly.  This works just fine on ARM
# because of the same issue, so allow it for now.
[ ${target} = "mips" ] || \
    HLINK_FILES="${HLINK_FILES} bin/sh bin/csh"
```

Use the one that Poudriere uses which will always be target.  _/usr/lib/crti1.o_
```
src/share/poudriere/jail.sh:        login_env="${login_env},ABI_FILE=\/usr\/lib\/crt1.o"
```

bdrewery:
Delaying this for 3.2.1.

PR #553, opened 2017-11-24
processonelog: add case for common lld linker error

emaste:
Needs to be before 'termios'.  Example error, from p5-Math-Pari:

/usr/bin/ld: error: can't create dynamic relocation R_AARCH64_ADR_PREL_PG_HI21 against symbol: gpmalloc in readonly segment
>>> defined in init.o
>>> referenced by mp.c
>>>               mp.o:(pari_kernel_init)


emaste:
Note I called this case `lld_linker_error`, because I think it will be useful to distinguish between errors specific to lld and existing linker errors, during the transition. But I would be happy to make it just `linker_error` if desired.

PR #554, opened 2017-11-25
(WIP) Package seeding

### LINKED ISSUE #319: Try to fetch dependencies instead of building them locally
Testing ports before committing changes is expensive. So, is building overlay of a few ports with non-default options.
- `/head` is a fast moving target, `ports -u` often renders many out-of-date
- Tier1 support consists of `4 * 2 = 8` platform combinations
- redports.org (CI) is still not available after it went down a year ago
- repo layout doesn't support `NO_ARCH`, `NO_OSREL`, `NO_OPSYS` yet
- some ports may require a lot of memory/time/diskactivity to build

Fortunately, pkg.freebsd.org already provides ready-to-use packages. Why not take advantage? Caveat: only useful when local ports tree and remote snapshot at the time don't diverge much.

yunchih:
Right after the first `clean_build_queue` call:

1. For all packages in `${MASTERMNT}/.p/all_pkgs`, if it's not marked "listed" or if it's not in
    `${PACKAGES}/All`, try fetch its `${pkg}-${ver}.txz` from pkg.freebsd.org.
    If it succeeds, place the .txz into `${PACKAGES}/All`.
2. Call `clean_build_queue` to remove fetched packages from build queue.

It is still not clear to me whether to place fetched packages in `${PACKAGES}/All` or in `${MASTERMNT}/packages/All`.  It seems that because `${MASTERMNT}/packages/All` is in jail, it is read-only, but that is where `clean_build_queue` checks.  Need help on this part.

Note: the code is still WIP and not working properly ...
Fix #319

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -6710,6 +6710,10 @@ prepare_ports() {
 
 	clean_build_queue
 
+	pkg_seeding
+
+	clean_build_queue
+
 	# Call the deadlock code as non-fatal which will check for cycles
 	msg "Sanity checking build queue"
 	bset status "sanity_check_queue:"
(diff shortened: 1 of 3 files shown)


yunchih:
Any update on this?  Any hint appreciated.

src/share/poudriere/common.sh
```diff
@@ -6877,6 +6881,64 @@ load_priorities() {
...
+
+	[ "${PWD}" = "${MASTERMNT}/.p" ] || \
+	    err 1 "pkg_seeding requires PWD=${MASTERMNT}/.p"
+
+	local url url_prefix jail_version jail_arch dest
+	local seed_total seed_ok
+
+	# Set up URL prefix
+	jail_version=$(jget ${JAILNAME} version)
+	# $MARJOR_RELEASE-$SUBRELEASE.RELEASE-px --> $MARJOR_RELEASE
```
dch: typo? s/MARJOR/MAJOR/ ?

src/share/poudriere/common.sh
```diff
@@ -6877,6 +6881,64 @@ load_priorities() {
...
+	fi
+
+	msg "Package seeding done, seeded ${seed_ok} packages."
+}
+
+pkg_seeding_download() {
+	[ $# -ne 3 ] && eargs pkg_seeding_download destination url originspec
+
+	msg_debug "Fetching seed from: $1"
+	fetch -q -a -w 1 -m -o "$2" "$1" &> /dev/null || \
```
dch: this could do with a comment; what value does the 2nd retry with different parameters bring here? does it solve an actual problem, and in the event of a failure can we rely on the retrieved package being correct? Should we check anywhere for validity, whether by SHA* or other method?

dch:
super idea - this is likely to help a lot of people move to poudriere for general port building.

I have one general question, if I read the code right, there is no way to detect if `make.conf` or custom config options for a dependent package (e.g. requiring an alternate SSL, or no docs, or some general version flag like `DEFAULT_VERSIONS+=pgsql=9.5` for example) would invalidate the requested package relying on seeded dependencies. You'd end up with a newer postgres (or perl or ruby ...) dependency being used to build against an older requested package.

Is there a mechanism elsewhere in poudriere (the code to detect changed ports/options) that can be borrowed to cater for this?

src/share/poudriere/common.sh
yunchih: Updated ( ;

src/share/poudriere/common.sh
yunchih: This kind of fetch retry is actually an imitation of the `fetch_file` function defined in the same file.  Its purpose is to recover from potential jitterish network that fails the first fetch.  However, we cannot verify the validity of fetched package without checking its SHA against pkg.freebsd.org's `digests.txz`.  As this will add quite a bit complexity, I will backlog it for now and finish the most basic functionality first. 😸

bdrewery:
It's a good start. It really needs to handle options, arch, versions, rather than blindly fetching and then possibly just deleting everything it just fetched. I'll bring it in and add the extra logic to handle that.

0mp:
This is probably going to be implement differently. See #797.

bdrewery:
Thank you for this. It's been superseded by #797 which uses `pkg`.

PR #569, opened 2017-12-19
processonelog: add case for explicit #error

emaste:
These are often things like '#error CPU arch not supported', and it's
convenient to have a specific message for these rather than ???


swills:
ok to test

Issue #586, opened 2018-01-23
ports tree creation quietly ignores branch by default


phryk:
`poudriere ports -c -p foo -B branches/2018Q1` defaults to poudriere using portsnap while quietly ignoring the supplied branch.

I might be a bit biased because I've been building HEAD for months while thinking I was on quarterlies, but I think it'd be best for this to explicitly fail with a descriptive message instructing the user to add '-m svn' or whatever is deemed best practice.

PR #587, opened 2018-01-23
made ports tree creation fail for portsnap when a branch was specified.

phryk:
This PR would fix issue #586, but it
- assumes that portsnap is the only (-m) method not supporting branches, which I haven't verified
- is a bit suboptimal because the check is only run after zfs dataset creation


bdrewery:
Thanks for the patch but I've done it a bit differently to capture all cases.

PR #603, opened 2018-03-08
Split poudriere.8 into subpages.

brooksdavis:
Break poudriere.8 into a manpage for each command similar to git(1).  The result is easier to search (try finding information about the `jail` or `ports` command with a text search in the old version) and each manpage is easier to read.

Discussed in passing with: @bdrewery


bdrewery:
This is awesome, thank you.

PR #604, opened 2018-03-09
Add a Bash completion for Poudriere

0mp:
I am not sure how to actually install this file, though. So for now I'm just putting it into the completions directory.

There are some rough edges here and there but it works well overall.

Suggestions welcome!

---

 * Version: 0.1.0:
   https://github.com/0mp/poudriere-bash-completion/releases/tag/0.1.0
 * Upstream: https://github.com/0mp/poudriere-bash-completion/

### DIFF
--- a/completions/bash/poudriere
+++ b/completions/bash/poudriere
@@ -27,15 +27,16 @@
 # SUCH DAMAGE.
 #
 # ---
-# Version: 0.1.0
+# Version: 0.2.0
+# Upstream: https://github.com/0mp/poudriere-bash-completion/
 
 _poudriere_absolute_path() {
     COMPREPLY=($(compgen -f "${cur:-/}" -- "$cur"))
 }
 
-# XXX: It might be non exhaustive.
+# XXX: It does not complete the architectures available thanks to the QEMU option.
 _poudriere_architecture() {
-    COMPREPLY=($(compgen -W 'amd64 i386' -- "$cur"))
+    COMPREPLY=($(compgen -W "$(sysctl -n kern.supported_archs)" -- "$cur"))
 }
 
 # TODO: Parse poudriere status.
@@ -47,6 +48,7 @@ _poudriere_direct_port() {
     # Find a ports tree to search for ports.
     local prev_word=
     local ports_tree=
+    local w
     # Find out if a ports tree is already specified with -p.
     for w in ${words[@]}; do
         if [[ $w == -p ]]; then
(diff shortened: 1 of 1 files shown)


smortex:
@0mp completions are to be installed by the port itself.  See [how it's done for zsh completions](https://svnweb.freebsd.org/ports/head/ports-mgmt/poudriere/Makefile?revision=459203&view=markup#l55) and the [“Shell Completion Files” section of the Porter's Handbook](https://www.freebsd.org/doc/en_US.ISO8859-1/books/porters-handbook/book.html#shell-completion) if you are interested in more info about this.

0mp:
@smortex Ohh right. It is so obvious now, thanks! 

It was late and I was confused by the fact that there is a `ZFS` option defined but on the other hand there is:

```Makefile
post-install:
	@${MKDIR} ${STAGEDIR}${PREFIX}/share/zsh/site-functions/
	${INSTALL_DATA} ${WRKSRC}/completions/zsh/_poudriere \
	    ${STAGEDIR}${PREFIX}/share/zsh/site-functions/
```

istead of

```Makefile
post-install-ZFS-on:
#...
```

I may submit a patch if you'd like to :)

smortex:
s/ZFS/ZSH/ 😉 

The completion is always staged, but added to the package only if the pkg-plist references it.  Nowadays, I would consider this being a not-so-good-practice but it should not hard too much, maybe this was done that way by the maintainer on purpose (ping @bdrewery). A few years back, it would have been a problem.

completions/bash/poudriere
```diff
@@ -0,0 +1,437 @@
...
+# ---
+# Version: 0.1.0
+
+_poudriere_absolute_path() {
+    COMPREPLY=($(compgen -f "${cur:-/}" -- "$cur"))
+}
+
+# XXX: It might be non exhaustive.
+_poudriere_architecture() {
+    COMPREPLY=($(compgen -W 'amd64 i386' -- "$cur"))
```
bdrewery: `sysctl -n kern.supported_archs` is a list of archs supported natively on this kernel.

completions/bash/poudriere
```diff
@@ -0,0 +1,437 @@
...
+
+_poudriere_filesystem() {
+    COMPREPLY=($(compgen -W "none $(zfs list | awk 'NR > 1{print $1}')" -- "$cur"))
+    if [[ ${#COMPREPLY[@]} == 1 ]] && [[ $COMPREPLY != none ]]; then
+        # XXX: Is it reasonable to assume that someone would continue to append somthing to an existing name?
+        compopt -o nospace
+    fi
+}
+
+_poudriere_freebsd_version() {
```
bdrewery: This is going to become bitrotted fast.

completions/bash/poudriere
```diff
@@ -0,0 +1,437 @@
...
+                            _poudriere_direct_port
+                        fi
+                        if [[ ${COMPREPLY[@]} -eq 0 ]]; then
+                            COMPREPLY=($(compgen -W "$(_parse_help poudriere $selected_command)" -- "$cur"))
+                        fi
+                        ;;
+                esac
+                ;;
+
+            # TODO: queue) ;;
```
bdrewery: Don't bother as the command is unstable and may go away.

completions/bash/poudriere
0mp: Fixed. I'll send an update soon.

completions/bash/poudriere
0mp: I've found a way to generate it dynamically :) Stay tuned.

completions/bash/poudriere
0mp: Done.

0mp:
@bdrewery I have not removed the version completion just yet as I found a way to prevent its bitrot. See [the commit message][1] and [the function comment][2] for more details.

[1]: https://github.com/0mp/poudriere-bash-completion/commit/be6a5f848e315378dd28e16ef2f244702ad850b4 
[2]: https://github.com/0mp/poudriere-bash-completion/blob/0.2.0/poudriere-bash-completion#L86-L96

PR #610, opened 2018-04-23
Use .Xr instead of .Cm to reference another manual

0mp:


bdrewery:
Thanks!

0mp:
@bdrewery Thank you too for maintaining this important tool!

PR #614, opened 2018-05-28
Resetting load attempts counter

TimonAlex:
The load attempts counter must be reset after successful loading of data.json otherwise code will react to total number of errors, not sequential one


bdrewery:
Thanks!

PR #615, opened 2018-06-03
Shush make by default in jail command

grimreaper:
poudriere can generate a lot of output, so try and keep it quieter in normal case

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -390,7 +390,7 @@ setup_build_env() {
 
 	JAIL_OSVERSION=$(awk '/^\#define[[:blank:]]__FreeBSD_version/ {print $3}' ${SRC_BASE}/sys/sys/param.h)
 	hostver=$(awk '/^\#define[[:blank:]]__FreeBSD_version/ {print $3}' /usr/include/sys/param.h)
-	MAKE_CMD=make
+	MAKE_CMD="make -s"
 	if [ ${hostver} -gt 1000000 -a ${JAIL_OSVERSION} -lt 1000000 ]; then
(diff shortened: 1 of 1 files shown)


mat813:
Why?

I don't really think this is a good idea.

grimreaper:
Most of the time when using poudriere you've already built the code at least once. Even if not, its quite common to using `-s` for the first build.

when doing tasks like:
```
[00:00:00] Copying /usr/src to /poudriere/jails/amd64-matches-host/usr/src... done
```
seeing all the `rm -f` and `install` commands is just noise. make(1) already tells you what its done if something went wrong.

bdrewery:
Make does not show the failing command on a build failure with `-s`. This is critical debugging information. Too bad we used `-v` in `jail` already or this could be gated on `VERBOSE` from the standard `-v`.

mat813:
What do you mean "you've already built the code at least once"?
I always build everything inside poudriere, so, if poudriere never shows me what it is doing, how can I see what is happening?

grimreaper:
> What do you mean "you've already built the code at least once"?
There are three possible cases:
(a) you've already built and now you're just installing (-b)
(b) you've already built, and you rebuild for poudriere (-b, but you have tested previously)
(c) the first time you're building the tree is for poudriere.

I suspect for most people (c) is far less common. Even when portmgr tests a patch, hopefully the provider of the patch has already tested it at least once.
For cases (a) and (b) poduriere produces excessive output which makes following the logs harder, and possibly slows down the update.

bdrewery:
As is this won't be taken as it is critical to see the build output somewhere for debugging failures.

PR #618, opened 2018-06-16
Add option to allow ZFS dataset access and extra parameters for builder jails

AndiDog:
This is the preparation for a use case I'm developing with @grembo and @xgarcias at our company PPRO. Namely, we want to package whole application jails (FreeBSD base + package dependencies + application) into a package for use in a new deployment mechanism. The presentation is proposed as talk "Joker: safe and automated deployments using FreeBSD jails" for EuroBSDcon 2018, and we would like to get the necessary patch ready so people could try the solution right after the talk instead of having to wait for a poudriere release. *I believe this patch is generally useful*, but also we are very sure to use this feature for our purposes (not a hobby project).

I can provide a sample port Makefile showing how it's used, but the basic idea is simple: instead of building a *software* port, we package a filesystem snapshot: `zfs create parentdataset/myappjail`, install FreeBSD release and packages in it, then ~ `zfs snapshot parentdataset/myappjail@export && zfs send parentdataset/myappjail@export | xz -c >zfs-image.txz`. We end up with a package containing that compressed ZFS dataset image which we can use in the port's post-install script to extract and make the jail ready once the package gets installed. The deployment mechanism then leverages safe switching between old and new jail whenever an application deployment needs to happen. If you want more info on this use case, let me know.

Or in summary: allow port builds to fiddle with a restricted list of ZFS datasets

-----------

Tested with 3.2.6, and then applied the patch cleanly to master branch.

### DIFF
--- a/src/etc/poudriere.conf.sample
+++ b/src/etc/poudriere.conf.sample
@@ -1,7 +1,6 @@
-
 # Poudriere can optionally use ZFS for its ports/jail storage. For
 # ZFS define ZPOOL, otherwise set NO_ZFS=yes
-# 
+#
 #### ZFS
 # The pool where poudriere will create all the filesystems it needs
 # poudriere will use ${ZPOOL}/${ZROOTFS} as its root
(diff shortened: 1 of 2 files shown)


bdrewery:
I'm starting to sour to adding things like this as it just increases complexity. People tend to think that *everything* in poudriere.conf needs to be handled. If they remove something like 'null' from devfs that's going to be a problem.

Can't you do this with a hook?

AndiDog:
Very good idea! Didn't even know that feature. I'll try to solve it as hook plugin, and if that works, close this PR.

AndiDog:
Works fine. Will open up a small documentation change PR instead.

PR #626, opened 2018-07-22
Ask for confirmation when cleaning all packages with 'bulk -c'

Jehops:
This would be helpful for morons like me who occasionally wipe all packages from
a repository with 'bulk -c' when 'bulk -C' was intended.  I think '-c' is in my
muscle memory for 'options -c'.


bdrewery:
This has turned out to be kind of annoying. I am going to make it only prompt if not using `ATOMIC_PACKAGE_REPOSITORY`. With that option then ` bulk -c` isn't the end of the world since you can just move some symlinks around to fix it.

PR #631, opened 2018-08-26
Use mkimg for rawdisk and zrawdisk to create valid partition tables

allanjude:
rawdisk and zrawdisk are meant to be full disk images, not just the contents of a partition.

booting ZFS from MBR sucks, use GPT.


allanjude:
It seems maybe these two types were not intended to have partition tables.

allanjude:
I misunderstood the point of 'rawdisk' mode. I'll propose a different change.

PR #632, opened 2018-08-26
Add -w (swap size) and -b (swap position)

allanjude:
Allow the user to specify how much swap, and if it should come before or after the primary partition

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -7729,31 +7729,45 @@ calculate_size_in_mb() {
 }
 
 calculate_ospart_size() {
+	local NUM_PART
 	local CALC_SIZE
 	local FULL_CALC_SIZE
 	local DATA_CALC_SIZE
+	local SWAP_CALC_SIZE
 	local CFG_CALC_SIZE
 
+	# How many partitions do we need
+	NUM_PART=$1
+
 	# Figure out the size of the image in MB
-	CALC_SIZE=${IMAGESIZE}
+	CALC_SIZE=$2
 	calculate_size_in_mb
 	FULL_CALC_SIZE=${CALC_SIZE}
 
 	# Figure out the size of the /cfg partition
-	CALC_SIZE=${CFG_SIZE}
+	CALC_SIZE=$3
 	calculate_size_in_mb
 	CFG_CALC_SIZE=${CALC_SIZE}
 
 	# Figure out the size of the Data partition
-	if [ $# -eq 3 -o -n ${DATA_SIZE} ]; then
-		CALC_SIZE=${DATA_SIZE}
+	if [ $# -ge 4 -o -n $4 ]; then
+		CALC_SIZE=$4
 		calculate_size_in_mb
 		DATA_CALC_SIZE=${CALC_SIZE}
 	else
 		DATA_CALC_SIZE=0
 	fi
 
-	OS_SIZE=$(( ( ${FULL_CALC_SIZE} - ${CFG_CALC_SIZE} - ${DATA_CALC_SIZE} ) / 2 ))
+	# Figure out the size of the swap partition
+	if [ $# -eq 5 -o -n $5 ]; then
+		CALC_SIZE=$5
+		calculate_size_in_mb
+		SWAP_CALC_SIZE=${CALC_SIZE}
+	else
+		SWAP_CALC_SIZE=0
+	fi
+
+	OS_SIZE=$(( ( ${FULL_CALC_SIZE} - ${CFG_CALC_SIZE} - ${DATA_CALC_SIZE} - ${SWAP_CALC_SIZE} ) / ${NUM_PART} ))
 	msg "OS Partiton size: ${OS_SIZE}m"
 }
 
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -7710,64 +7710,60 @@ build_repo() {
 }
 
 calculate_size_in_mb() {
-	case ${CALC_SIZE} in
+	local calc_var="$1"
+	local calc_size
+
+	getvar "$calc_var" calc_size
+
+	case ${calc_size} in
 	*p)
-		CALC_SIZE=${CALC_SIZE%p}
-		CALC_SIZE=$(( ${CALC_SIZE} << 10 ))
+		calc_size=${calc_size%p}
+		calc_size=$(( ${calc_size} << 10 ))
 		;&
 	*t)
-		CALC_SIZE=${CALC_SIZE%t}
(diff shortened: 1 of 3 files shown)


bapt:
What if I don't want any swap?

allanjude:
Good point, I should add an if for handling 0 better

bapt:
The PR is now almost 1year old to you plan to provide an updated version ?

allanjude:
I do, sorry I've been so busy.

allanjude:
Burried deep in the couch cushions, I found a "roundtoit"

allanjude:
Rebased to 2020-05-08

src/share/poudriere/common.sh
```diff
@@ -7729,31 +7729,45 @@ calculate_size_in_mb() {
...
 	# Figure out the size of the /cfg partition
-	CALC_SIZE=${CFG_SIZE}
+	CALC_SIZE=$3
 	calculate_size_in_mb
 	CFG_CALC_SIZE=${CALC_SIZE}
 
 	# Figure out the size of the Data partition
-	if [ $# -eq 3 -o -n ${DATA_SIZE} ]; then
-		CALC_SIZE=${DATA_SIZE}
+	if [ $# -ge 4 -o -n $4 ]; then
```
bdrewery, line 7753: I think this pattern can be error prone and it is best to resolve all params up front like:
```
local NUM_PART="$1"
local IMAGESIZE="$2"
local CFG_SIZE="$3"
local DATA_SIZE="$4"
local SWAP_CALC_SIZE="$5"
```
...
```
if [ -n "$DATA_SIZE" ]; then
CALC_SIZE="$DATA_SIZE"
...
fi
```
...
```
if [ -n "$SWAP_CALC_SIZE" ]; then
CALC_SIZE="$SWAP_CALC_SIZE"
...
fi
```

Also having said all of that I don't know why `CALC_SIZE` is a thing. `calculate_size_in_mb` should take a proper parameter. It can return the value wanted directly.
```
calculate_size_in_mb() {                       
        local CALC_SIZE="$1"
        local var_return="$2"

        case ${CALC_SIZE} in                   
        *p)                                    
                CALC_SIZE=${CALC_SIZE%p}       
                CALC_SIZE=$((CALC_SIZE << 10 ))
                ;&                             
        *t)                                    
                CALC_SIZE=${CALC_SIZE%t}       
                CALC_SIZE=$((CALC_SIZE << 10 ))
                ;&                             
        *g)                                    
                CALC_SIZE=${CALC_SIZE%g}       
                CALC_SIZE=$((CALC_SIZE << 10 ))
                ;&                             
        *m)                                    
                CALC_SIZE=${CALC_SIZE%m}       
        esac 
       
        setvar "$var_return" "$CALC_SIZE"                                  
}                                              
```

Then called like:
```
calculate_size_in_mb "$SWAP_CALC_SIZE" SWAP_CALC_SIZE
```
It would be easy to make it `calculate_size_in_mb SWAP_CALC_SIZE` too if that is more appealing:
```
calculate_size_in_mb() {                       
        local calc_var="$1"
        local CALC_SIZE

        getvar "$calc_var" CALC_SIZE

        case ${CALC_SIZE} in                   
        *p)                                    
                CALC_SIZE=${CALC_SIZE%p}       
                CALC_SIZE=$((CALC_SIZE << 10 ))
                ;&                             
        *t)                                    
                CALC_SIZE=${CALC_SIZE%t}       
                CALC_SIZE=$((CALC_SIZE << 10 ))
                ;&                             
        *g)                                    
                CALC_SIZE=${CALC_SIZE%g}       
                CALC_SIZE=$((CALC_SIZE << 10 ))
                ;&                             
        *m)                                    
                CALC_SIZE=${CALC_SIZE%m}       
        esac 
       
        setvar "$calc_var" "$CALC_SIZE"                                  
}  
```
...
```
calculate_size_in_mb SWAP_CALC_SIZE
```

src/share/poudriere/common.sh
```diff
@@ -7729,31 +7729,45 @@ calculate_size_in_mb() {
...
+	# Figure out the size of the swap partition
+	if [ $# -eq 5 -o -n $5 ]; then
+		CALC_SIZE=$5
+		calculate_size_in_mb
+		SWAP_CALC_SIZE=${CALC_SIZE}
+	else
+		SWAP_CALC_SIZE=0
+	fi
+
+	OS_SIZE=$(( ( ${FULL_CALC_SIZE} - ${CFG_CALC_SIZE} - ${DATA_CALC_SIZE} - ${SWAP_CALC_SIZE} ) / ${NUM_PART} ))
```
bdrewery, line 7770: So incidentally I made changes to this line yesterday but have not pushed to remove the `${}` as they are not needed in `$(( ))``.
I will remove the change from my branch.

I also had moved both this function and `calculate_size_in_mb` out of common.sh and into image.sh since they are not needed by any other file. I'll revert this in my local change as well for now.

src/share/poudriere/common.sh
bdrewery: For reference the `getvar` is just a wrapper around an `eval` but `setvar` is a FreeBSD sh extension.

bdrewery: In case my edit is lost in the noise my last `calculate_size_in_mb` has a bug.
```
### Can't use CALC_SIZE in this function for it to work or the setvar just sets the
### local version of the variable. I didn't have a real editor for this but a simple
### s/CALC_SIZE/calc_size/ in the function would avoid the problem.
```

allanjude:
Thanks for the extensive feedback

src/share/poudriere/common.sh
allanjude: Yeah, that works out much better

src/share/poudriere/common.sh
allanjude: Do you want me to move these 2 functions to the image.sh file?

src/share/poudriere/common.sh
```diff
@@ -7710,50 +7710,60 @@ build_repo() {
...
+
+	if [ -n "CFG_SIZE" ]; then
+		calculate_size_in_mb CFG_SIZE
 	else
-		DATA_CALC_SIZE=0
+		CFG_SIZE=0
 	fi
-
-	OS_SIZE=$(( ( ${FULL_CALC_SIZE} - ${CFG_CALC_SIZE} - ${DATA_CALC_SIZE} ) / 2 ))
+	if [ -n "DATA_SIZE" ]; then
```
bdrewery, line 7755: er `$DATA_SIZE`... I'll fix

allanjude: Sorry I missed that

PR #647, opened 2018-10-17
image: adds zsnapshot imagetype

baloo:
The purpose of this format is to be able to run jails off a zfs snapshot
a couple of files would be generated:
  - NAME-VERSION.manifest.json
  - NAME-VERSION.full.img.gz
  - NAME-latest.manifest.json -> NAME-VERSION.manifest.json (symlink)
and optionaly (if image is generated off a previous snapshot):
  - NAME-VERSION.incr.img.gz
  - NAME-VERSION.modified.files

The full image is the result of zfs send ran off the image builder,
the incremental zfs image is a diff from previous version.

No kernel is present on those images, and the purpose is for servers
to download those files by however means (http, ftp, scp, ...),
verify the files hash, and import them on their local zpool.

### DIFF
--- a/src/share/poudriere/image.sh
+++ b/src/share/poudriere/image.sh
@@ -34,6 +34,7 @@ Parameters:
                        the image
     -f packagelist  -- List of packages to install
     -h hostname     -- The image hostname
+    -i originimage  -- Origin image name
     -j jail         -- Jail
     -m overlaydir   -- Build a miniroot image as well (for tar type images), and
                        overlay this directory into the miniroot image
@@ -41,10 +42,11 @@ Parameters:
     -o outputdir    -- Image destination directory
     -p portstree    -- Ports tree
     -s size         -- Set the image size
+    -S snapshotname -- Snapshot name
     -t type         -- Type of image can be one of (default iso+zmfs):
                     -- iso, iso+mfs, iso+zmfs, usb, usb+mfs, usb+zmfs,
                        rawdisk, zrawdisk, tar, firmware, rawfirmware,
-                       embedded, dump
+                       embedded, dump, zsnapshot
     -X excludefile  -- File containing the list in cpdup format
     -z set          -- Set
 EOF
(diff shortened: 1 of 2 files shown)


baloo:
Please do not merge this PR just yet, I'm opening it early to gather comments on the idea.

baloo:
I'm also adding a dependency on rsync. The purpose here is to get the smallest incremental snapshot as possible, rewriting from scratch the filesystem off tarball images makes zfs create new objects. I do not want that, so I'm running rsync instead to only rewrite files that changed (most does not across builds).

src/share/poudriere/image.sh
```diff
@@ -334,6 +345,30 @@ zrawdisk)
...
 	zfs create -o mountpoint=/var/empty ${zroot}/var/empty
 	;;
+zsnapshot)
+	zfs list ${ZPOOL}${ZROOTFS}/images >/dev/null 2>/dev/null || \
+		zfs create -o compression=lz4 ${ZPOOL}${ZROOTFS}/images
+	zfs list ${ZPOOL}${ZROOTFS}/images/work >/dev/null 2>/dev/null || \
+		zfs create ${ZPOOL}${ZROOTFS}/images/work
+	mkdir -p ${WRKDIR}/mnt
+	if [ ! -z "${ORIGIN_IMAGE}" ]; then
+		cat ${ORIGIN_IMAGE} | \
```
bapt: The cat call can be removed

src/share/poudriere/image.sh
```diff
@@ -503,6 +538,9 @@ tar)
 		mkminiroot
 	fi
 	;;
+zsnapshot)
+	rsync -ac --delete ${WRKDIR}/world ${WRKDIR}/mnt
```
bapt: poudriere bundles cpdup and put it upfront in the path, so you can use it instead of rsync :)

baloo:
@bapt is there anything else I can do?

bapt:
merged by another mean (resolving conflict) sorry for the delay

baloo:
@bapt thanks!

PR #648, opened 2018-10-18
jails, git: remote may have updated branch non-fastfoward

baloo:
poudriere ports is expected to be the same as remote. If remote has
updated the objects in non-fast-forward mode (rebase, git push force, ...).
The pull will try to merge objects and will leave dirty merge markers.

This commits removes this behavior and tries to make the local clone
build the exact same objects as those in the remote repository.

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -582,7 +582,9 @@ install_from_vcs() {
 			echo " done"
 			;;
 		git*)
-			${GIT_CMD} -C ${SRC_BASE} pull --rebase -q || err 1 " fail"
+			${GIT_CMD} -C ${SRC_BASE} fetch -q || err 1 " fail"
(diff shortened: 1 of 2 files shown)


bdrewery:
Can you handle this one too please?
```
src/share/poudriere/jail.sh:                    ${GIT_CMD} -C ${SRC_BASE} pull --rebase -q || err 1 " fail"
```

baloo:
@bdrewery there!

bdrewery:
Sigh I wonder if this change will surprise people. Both the removal of `rebase` and the addition of `git clean`. I'm sure people have random patches sitting in their trees. And I have to wonder if people have committed to their local checkout knowing that `git pull --rebase` just works.
I think I'm more concerned about the `git clean` addition as I know people drop patches in (I do it), and the other update methods (SVN) don't do similar.

baloo:
@bdrewery what would you think of a simple git status before update (and abort if jail is not clean)?

Issue #592, opened 2018-02-06
Feature request: non-interactive deletion of ports collections, jails etc.


bvdwiel:
Trying to manage Poudriere through Puppet. My module is progressing nicely, but removal of ports collections etc. is impossible due to the command throwing a Y/n prompt which Puppet can't handle since it's non-interactive. It'd be nice to have a command-line flag to always assume 'Y' so that all commands can be executed by non-interactive tools.

bdrewery:
What command? pkgclean?
```
    -n          -- Do not actually remove anything, just show what would be
                   removed
    -y          -- Assume yes when deleting and do not confirm
```

bvdwiel:
What I'm looking for, is a way to do something like:
 poudriere ports -d -p local

..but without needing to answer Y to a prompt. Maybe I'm missing something here, in that case I'll submit a PR to the docs ;-)

bdrewery:
`ports -d` already only prompts if stdin is a tty, otherwise it assumes yes. There is also always `yes | poudriere`.

bdrewery:
Can you confirm whether or not puppet itself got stuck on the prompt, or just your testing outside of puppet?

bvdwiel:
I tested actual Puppet runs and those do work. I was thrown off by the prompts I get in a TTY and didn't expect different behaviour. So if anything, this difference might be addressed in the docs a little more explicitly.

brd:
So should we add a `-y` to `poudriere ports` just like I added to `poudriere jail` in ae7bc3723f29e862c89b51a97ef6c89a12453c5d, or is that not needed?

bdrewery:
I guess it makes sense for consistency. Go for it if you want. I don't think it is needed as it only prompts in a TTY and `yes` is a thing, I meant to close this when #649 went in so closing now.

devosalain:
/usr/local/bin/poudriere ports -k -d                       -p $PORTS
prompts even if it is in a script , 
this is minor ,
but annoying.

PR #649, opened 2018-10-22
Document the confirmation prompt for "jail -d".

0mp:
PR:	#592


0mp:
I guess we should update all the other poudriere commands(like *ports*) as well, right?

PR #652, opened 2018-11-07
zsh completions: wip to complete specific ports in a tree

Jehops:
There is one annoyance.  For options that expect an argument, if the user hits a space before trying to complete, zsh completes on both the argument and a ports tree.  For example, if the user enters `poudriere bulk -j` then hits `TAB`, then zsh will only complete on jails, but if the user enters `poudriere bulk -j `, then zsh will complete both the jails and ports.  Other than that all seems to be working well.  If you do consider this useful, it might be best to wait a few days, so I can test more.  I open the PR now, in case there are any changes to discuss.


Jehops:
So far, so good.  It seems to work well.

bdrewery:
Can you explain in your commit, or split into two, the various changes beyond the ports tree change? Also your commit still says "wip".

Jehops:
I updated the commit and included a description of the two types of changes in the commit log. 

1. The first change ensures that mutually exclusive options are completed properly. For example, if `-a` is included with `poudriere bulk`, then `-f` or a ports tree should not be completed.

2. Specifying mandatory option arguments correctly fixes the annoyance that I described in the original comment.

bdrewery:
Hmm now it seems completions don't require any `-` flags. Like testport still needs a `-j` but it jumps right to ports:
```
~/git/poudriere3 # poudriere testport
Keywords/       accessibility/  audio/          cad/            databases/      dns/            french/         graphics/       japanese/       mail/           net/            news/           ports-mgmt/     science/        textproc/       x11/            x11-fonts/      x11-wm/
Mk/             arabic/         base/           chinese/        deskutils/      editors/        ftp/            hebrew/         java/           math/           net-im/         packages/       portuguese/     security/       ukrainian/      x11-clocks/     x11-servers/   
Templates/      archivers/      benchmarks/     comms/          devel/          emulators/      games/          hungarian/      korean/         misc/           net-mgmt/       palm/           print/          shells/         vietnamese/     x11-drivers/    x11-themes/
Tools/          astro/          biology/        converters/     distfiles/      finance/        german/         irc/            lang/           multimedia/     net-p2p/        polish/         russian/        sysutils/       www/            x11-fm/         x11-toolkits/
```

bdrewery:
And using `-p <other>` didn't grab the ports and even ran `poudriere`
```
~/git/poudriere3 # poudriere bulk -p commit [00:00:00] Error: Signal SIGINT caught, cleaning up and exiting
                   poudriere bulk -p commit fg
[00:00:00] Error: Don't know on which jail to run please specify -j
```
(I didn't hit enter at all there)

bdrewery:
What's the sane way to reload the completions when replacing the `_poudriere` file?

Jehops:
> Hmm now it seems completions don't require any `-` flags. Like testport still needs a `-j` but it jumps right to ports:

Sorry, I didn't catch that `poudriere testport cat/port -j <jail>` was not valid syntax. For `testport` if `-o` is not specified than I will ensure that `-j <jail>` is specified before `cat/part`. I will also take another look at the other commands.

Note that if you hit a `-` first, then the `-*` options are completed.

bdrewery:
The `-j` problem is on `bulk` too. In general the ports are expected to come last. The `-o` flag for `testport` is not even needed. It's an accident of history.

Jehops:
> What's the sane way to reload the completions when replacing the `_poudriere` file?

`rehash` seems to work. Apparently this only works if you haven't unset the `hash_list_all` option though.

I hope to take another go at this soon.

PR #653, opened 2018-11-16
poudriere image: Remove comments when passing contents of package lis…

Savagedlight:
…t file to "pkg add".

Sponsored by: Modirum


Savagedlight:
Will submit a better fix for this in a while.

PR #658, opened 2018-12-21
Display the progress indicator extracted from build logs

yurivict:


bdrewery:
Looks interesting. Can you add an example into the PR please?

yurivict:
@bdrewery For example, you can build ```net-im/tox``` or ```net-im/qTox```, all ```cmake```-based projects print progress indicators.

bdrewery:
> @bdrewery For example, you can build `net-im/tox` or `net-im/qTox`, all `cmake`-based projects print progress indicators.

I could but my time is limited. Can you copy/paste or screenshot the output please?

yurivict:
Attaching the screenshot:

image

bdrewery:
Oh very cool. It seems like it might go really wrong on non-cmake things though?

yurivict:
>  It seems like it might go really wrong on non-cmake things though?

No, it looks at a certain number of last lines in the log, and if the tags aren't there it produces a blank output.

bdrewery:
Thanks!

bdrewery:
```
23:39:28 <@mat> bdrewery, if I hit c^T, I get:
23:39:28 <@        mat>| [11amd64-pkgng-pg96] [20190227T082046Z] [parallel_build:] Queued: 56 Built: 7  Failed: 13 Skipped: 22 Ignored: 10 Tobuild: 4   Time: 00:18:01
23:39:31 <@        mat>| sed: 1: "s,^\[( *[0-9]+%|[0-9]+/ ...": RE error: repetition-operator operand invalid
23:39:34 <@        mat>| tail: stdout
23:39:37 <@        mat>|         [03]: graphics/gdal             | gdal-2.4.0                build            (00:14:54 / 00:16:16)
```
@mat813

bdrewery:
```
# sed -E "s,^\[( *[0-9]+%|[0-9]+/[0-9]+)\].*,{progress}\1,; /^{progress}/date; s/{progress}//; s/ //g"
sed: 1: "s,^\[( *[0-9]+%|[0-9]+/ ...": RE error: repetition-operator operand invalid
# gsed -E "s,^\[( *[0-9]+%|[0-9]+/[0-9]+)\].*,{progress}\1,; /^{progress}/date; s/{progress}//; s/ //g"
gsed: -e expression #1, char 63: Invalid preceding regular expression
```

Grr this doesn't work at all.

bdrewery:
Changing `{progress}` to `#progress#` seems to work. It must be trying to use the `{}` for repetition. Why does 11/head consider it invalid but your testing did not?
I have to revert for now.

bdrewery:
`regex(3)` says:
```
A ‘{’ followed by a digit is considered the beginning of bounds for a bounded repetition, which must then follow the syntax for bounds.  A ‘{’ not followed by a digit is considered an ordinary character.
```
So `sed` is broken?

yurivict:
I didn't see this at all. I am running 3.2.99.20190225 on 11.2 STABLE amd64.
Investigating.

bdrewery:
Ok I'll wait and not revert for now. Will need to do a 3.0.1.

yurivict:
Hmm, I can't reproduce it with ```graphics/gdal```, or with any other port. I was running with this patch for a month. Wondering what the problem might be.

bdrewery:
Perhaps it is r322211
```
commit 39d016e70fde91c131d5276cce9b4bbd4ee71259
Author: kevans <kevans@FreeBSD.org>
Date:   Tue Aug 8 04:10:46 2017 +0000

    regex(3): Handle invalid {} constructs consistently and adjust tests

    Currently, regex(3) exhibits the following wrong behavior as demonstrated
    with sed:

     - echo "a{1,2,3}b" | sed -r "s/{/_/"     (1)
     - echo "a{1,2,3}b" | sed "s/\}/_/"       (2)
     - echo "a{1,2,3}b" | sed -r "s/{}/_/"    (3)

    Cases (1) and (3) should throw errors but they actually succeed, and (2)
    throws an error when it should match the literal '}'. The correct behavior
    was decided by comparing to the behavior with the equivalent BRE (1)(3) or
    ERE (2) and consulting POSIX, along with some reasonable evaluation.

    Tests were also adjusted/added accordingly.

    PR:             166861
    Reviewed by:    emaste, ngie, pfg
    Approved by:    emaste (mentor)
    MFC after:      never
    Differential Revision:  https://reviews.freebsd.org/D10315

Notes:
    svn path=/head/; revision=322211

```
```
@@ -438,9 +439,6 @@ p_ere_exp(struct parse *p, struct branchc *bc)
                        break;
                }
                break;
-       case '{':               /* okay as ordinary except if digit follows */
-               (void)REQUIRE(!MORE() || !isdigit((uch)PEEK()), REG_BADRPT);
-               /* FALLTHROUGH */
        default:
```
The case for non-digit was lost.
@kevans91

bdrewery:
Anyway I'm testing with `#progress#` and will release that if it works.

yurivict:
I think you are right, this because of this sed change.
Please replace ```{progress}``` with ```#progress#``` or any thing else "special"-looking. This is just a placeholder.

kevans91:
Unfortunately, this was UB territory by POSIX standards (section 9.4.3 of IEEE 1003.1-2017 [1]) that I took the completely opposite direction of how we've historically handled it, and I failed to update documentation to boot. I'll rectify this to match the documentation, because I don't have any strong feelings about it.

[1] http://pubs.opengroup.org/onlinepubs/9699919799/

bdrewery:
Sorry but I'm going to revert this for now.
See https://github.com/freebsd/poudriere/issues/671#issuecomment-469418878 for my reasoning.
I would have added an option to use it but it won't be compatible with my plans for distributed. So it will have to come out eventually anyway.

Issue #654, opened 2018-11-20
Device busy after build


royger:
Hello,

While doing a bulk build of packages with poudriere I got the following error:

```
2018-11-19 17:55:38 Z executing ssh ... root@172.16.144.43 poudriere bulk -j builder -p local -f ~/package.list
 
[00:00:00] Creating the reference jail... done
[00:00:01] Mounting system devices for builder-local
[00:00:01] Mounting ports/packages/distfiles
[00:00:01] Converting package repository to new format
[00:00:01] Stashing existing package repository
[00:00:01] Mounting packages from: /home/osstest/build.130518.build-amd64-freebsd-packages/poudriere/packages/builder-local
/etc/resolv.conf -> /home/osstest/build.130518.build-amd64-freebsd-packages/poudriere/.m/builder-local/ref/etc/resolv.conf
[00:00:01] Starting jail builder-local
[00:00:02] Logs: /home/osstest/build.130518.build-amd64-freebsd-packages/poudriere/logs/bulk/builder-local/2018-11-19_17h55m34s
[00:00:02] Loading MOVED for /home/osstest/build.130518.build-amd64-freebsd-packages/poudriere/.m/builder-local/ref/usr/ports
[00:00:04] Ports supports: FLAVORS SELECTED_OPTIONS
[00:00:04] Gathering ports metadata
[00:00:06] Calculating ports order and dependencies
[00:00:06] pkg package missing, skipping sanity
[00:00:06] Skipping incremental rebuild and repository sanity checks
[00:00:06] Cleaning the build queue
[00:00:06] Sanity checking build queue
[00:00:06] Processing PRIORITY_BOOST
[00:00:06] Balancing pool
[00:00:06] Recording filesystem state for prepkg... done
[00:00:06] Building 116 packages using 6 builders
[00:00:06] Starting/Cloning builders
[00:00:07] Hit CTRL+t at any time to see build progress and stats
[00:00:07] [01] [00:00:00] Building ports-mgmt/pkg | pkg-1.10.5_5
[00:01:55] [01] [00:01:48] Finished ports-mgmt/pkg | pkg-1.10.5_5: Success
[00:01:55] [01] [00:00:00] Building print/indexinfo | indexinfo-0.3.1
[00:01:55] [02] [00:00:00] Building lang/perl5.26 | perl5-5.26.2_2
[00:01:55] [03] [00:00:00] Building devel/pkgconf | pkgconf-1.5.4,1
[...]

PR #661, opened 2019-02-12
fs: fix unmount of directories in destroyfs

royger:
Current usage of umountfs by destroyfs is wrong. umountfs is currently
called with childonly set from destroyfs, which means that the path
provided in the first argument won't be unmounted, only the
mount points contained within. The code in destroyfs that follows the
call to umountfs will try to remove the directory passed to umountfs,
but that's likely not possible since the directory itself can be a
mount point.

Fix this by calling umountfs without childonly, and thus make sure the
path provided in the first parameter is unmounted.

Fixes: https://github.com/freebsd/poudriere/issues/654

### DIFF
--- a/src/share/poudriere/include/fs.sh
+++ b/src/share/poudriere/include/fs.sh
@@ -349,7 +349,7 @@ destroyfs() {
 	local type="$2"
 	local fs
 
-	umountfs ${mnt} 1
+	umountfs ${mnt}
 	if [ ${TMPFS_ALL} -eq 1 ]; then
 		if [ -d "${mnt}" ]; then
 			if ! umount ${UMOUNT_NONBUSY} "${mnt}" 2>/dev/null; then


src/share/poudriere/include/fs.sh
```diff
@@ -349,7 +349,7 @@ destroyfs() {
 	local type="$2"
 	local fs
 
-	umountfs ${mnt} 1
+	umountfs ${mnt}
 	if [ ${TMPFS_ALL} -eq 1 ]; then
 		if [ -d "${mnt}" ]; then
 			if ! umount ${UMOUNT_NONBUSY} "${mnt}" 2>/dev/null; then
```
bdrewery: This unmounts the main mount for tmpfs.
Line 362 unmounts the main mount for zfs.
It seems that `NO_ZFS` may not unmount the main mount. Is that the configuration you are using?

bdrewery:
Thanks for your report, analysis and patch. Fixed over in #654 differently.

PR #664, opened 2019-02-24
Base Ports

kmoore134:
This PR adds support for TrueOS-inspired "base-ports" for package base installations.

Namely it does the following:

- Add new "-m ports=PORTSTREE" option to jail creation METHOD.

This allows building a poudriere jail from the base ports already included in a checked out portstree, similar to how src=SRCDIR method operates. At the end, packages are created and stashed away to be included in final package outputs later. 

- Add new "-m pkg=URL" option to jail creation METHOD.

This allows using an existing base-packages pkg repo online (or local with file://) to boot-strap a jail, avoiding the need for dist-files. 

For those who wish to look at our ports structure, here's a [link](https://github.com/trueos/trueos-ports/tree/trueos-master/os) to the new OS category were currently using with this patch.


src/share/poudriere/jail.sh
```diff
@@ -612,6 +712,47 @@ install_from_vcs() {
...
+	INSLIST="os/userland os/src os/kernel os/buildworld os/buildkernel"
+
+	# Make pkg play nice
+	export IGNORE_OSVERSION="YES"
+
+	for inspkg in ${INSLIST}
+	do
+		msg "Installing ${inspkg} into jail..."
+		# Install the packages
+		pkg-static -r ${JAILMNT} -o ABI="${ABISTRING}" \
```
bdrewery: `pkg-static` here but `pkg` above?

src/share/poudriere/jail.sh
```diff
@@ -612,6 +712,47 @@ install_from_vcs() {
...
+
+	cat >${JAILMNT}/pkgrepo/repo.conf <<EOF
+base: {
+  url: "${PKGREPO}",
+  signature_type: "none",
+  enabled: yes,
+}
+EOF
+	# Get the remote version
+	REMOTEVER=$(sudo pkg rquery "%At=%Av" ports-mgmt/pkg | grep "FreeBSD_version" | cut -d '=' -f 2 | cut -c 1-2)
```
bdrewery: Why is `sudo` needed for `pkg rquery`?

src/share/poudriere/bulk.sh
```diff
@@ -219,6 +219,8 @@ if [ -d ${LOGD} -a ${CLEAN} -eq 1 ]; then
 	[ ${DRY_RUN} -eq 0 ] && rm -Rf ${LOGD} 2>/dev/null
 fi
 
+ingest_base_pkgs
+
 prepare_ports
```
bdrewery: `ingest_base_pkgs` is called before `prepare_ports` in both commands so it should be in there. You can use `if was_a_bulk_run` if needed to limit it to package builds.

src/share/poudriere/common.sh
```diff
@@ -7477,6 +7477,24 @@ calculate_ospart_size() {
 	msg "OS Partiton size: ${OS_SIZE}m"
 }
 
+ingest_base_pkgs()
+{
+	# Do we have any base-port packages to ingest?
+	if [ ! -d "${MASTERMNT}/.packages/" ] ; then
```
bdrewery: Style, space before `;` isn't normal here.

src/share/poudriere/common.sh
```diff
@@ -7477,6 +7477,24 @@ calculate_ospart_size() {
...
+		return 0
+	fi
+
+	if [ ! -d "${PACKAGES}/All/" ] ; then
+		mkdir -p "${PACKAGES}/All/"
+	fi
+
+	for i in $(ls ${MASTERMNT}/.packages/*.txz)
```
bdrewery: No reason to execute `ls`. You can avoid empty dir with a `case` as done elsewhere.

src/share/poudriere/jail.sh
```diff
@@ -57,8 +57,8 @@ Options:
     -m method     -- When used with -c, overrides the default method for
                      obtaining and building the jail. See poudriere(8) for more
                      details. Can be one of:
-                       allbsd, ftp-archive, ftp, git, http, null, src=PATH, svn,
-                       svn+file, svn+http, svn+https, svn+ssh, tar=PATH
+                       allbsd, ftp-archive, ftp, git, http, null, pkg=URL, ports=PATH,
+		       src=PATH, svn, svn+file, svn+http, svn+https, svn+ssh, tar=PATH,
```
bdrewery: Might as well fix the whitespace issue here

src/share/poudriere/jail.sh
```diff
@@ -612,6 +712,47 @@ install_from_vcs() {
 	    ${JAILMNT}/usr/include/sys/param.h)"
 }
 
+install_from_pkg() {
+	mkdir ${JAILMNT}/pkgrepo
+	local ABISTRING INSLIST REMOTEVER
+
+	cat >${JAILMNT}/pkgrepo/repo.conf <<EOF
```
bdrewery: Using `-EOF` will allow tab-indenting everything.

src/share/poudriere/jail.sh
```diff
@@ -505,6 +516,95 @@ build_native_xtools() {
...
+	make -C ${PORTS_BASE}/os/src WRKDIR=${JAILMNT}/work/src BATCH=yes package >${LOGDIR}/${JAILNAME}-buildsrc.log 2>&1
+	if [ $? -ne 0 ] ; then
+		err 1 "Failed building src"
+	fi
+
+	# Install the package
+	local PKGFILE="${JAILMNT}/work/src/pkg/$(make PORTSDIR=${PORTS_BASE} -C ${PORTS_BASE}/os/src -V PKGNAME).txz"
+	local ABISTRING="$(make PORTSDIR=${PORTS_BASE} -C ${PORTS_BASE}/os/src -V PKG_ABISTRING)"
+	pkg-static -o ABI=${ABISTRING} -r ${JAILMNT} add ${PKGFILE}
+	if [ $? -ne 0 ] ; then
```
bdrewery: This style of checking `$?` isn't right in Poudriere, nor needed.
`if ! pkg-static ...; then
    err
fi
`
or simply
`pkg-static ... || err`

src/share/poudriere/jail.sh
```diff
@@ -846,6 +987,22 @@ create_jail() {
...
+		PKGREPO="${METHOD##*=}"
+		[ -z "${PKGREPO}" ] && \
+		    err 1 "Must use format -m pkg=<url>/<to>/<repo>"
+		METHOD="${METHOD%%=*}"
+		;;
+	ports=*)
+		PTNAME="${METHOD#ports=}"
+	        # test if it already exists
+		porttree_exists ${PTNAME} || err 1 "No such ports name ${PTNAME}"
+		PORTS_BASE="${BASEFS:=/usr/local${ZROOTFS}}/ports/${PTNAME}"
```
bdrewery: The hardcoded paths here seem to be making a lot of assumptions.

src/share/poudriere/jail.sh
```diff
@@ -846,6 +987,22 @@ create_jail() {
...
+		test -d ${PORTS_BASE} || err 1 "No such ports directory"
```
bdrewery: No need to call `test` and not the right style. `[ -d "${PORTS_BASE}" ] || err`

src/share/poudriere/jail.sh
```diff
@@ -505,6 +516,95 @@ build_native_xtools() {
...
+
+		# Copy the package to the repo
+		msg "Copying package files"
+		cp ${PKGFILE} ${PACKAGES}/
+		if [ $? -ne 0 ] ; then
+			err 1 "Failed copying package"
+		fi
+
+		msg "Cleaning port files"
+		make -C ${PORTS_BASE}/os/build${tgt} WRKDIR=${JAILMNT}/work/${tgt} SRCDIR=${JAILMNT}/usr/src BATCH=yes clean >/dev/null 2>/dev/null
```
bdrewery: `make clean` can fail

src/share/poudriere/jail.sh
```diff
@@ -505,6 +516,95 @@ build_native_xtools() {
...
+	msg "Copying package files"
+	cp ${PKGFILE} ${PACKAGES}/
+	if [ $? -ne 0 ] ; then
+		err 1 "Failed copying package"
+	fi
+
+	# Cleanup the src package
+	msg "Cleaning port files"
+	make -C ${PORTS_BASE}/os/src WRKDIR=${JAILMNT}/work/src BATCH=yes clean 2>/dev/null >/dev/null
+	make -C ${PORTS_BASE}/os/src WRKDIR=${JAILMNT}/work/src BATCH=yes distclean 2>/dev/null >/dev/null
```
bdrewery: These can fail

src/share/poudriere/jail.sh
bdrewery: It's also wrong in Poudriere because the use of `set -e` will never make it to the `if [ $?` line. `set -e` might be off in this file but that's already a bug.

src/share/poudriere/jail.sh
kmoore134: This is fixed now, however I just pulled this style from the src= section a few lines below:

`test -d ${SRC_BASE} || err 1 "No such source directory"`

FYI in case you want to fix that one also later.

kmoore134:
> I feel like this should all just be called from hooks. Perhaps more hooks are needed.
> I don't, and can't, maintain things like this for systems and trees I don't have.

Understood. I feel these changes are small enough and non-intrusive that it shouldn't cause you additional maintenance work for existing functionality. And of course we will be using them for FreeNAS and keep them functioning for the foreseeable future. Assuming you merge, I'll be happy if you assign issues related to this over to me going forward. 

I'd like to push this style of base packages for non-TrueOS platforms also, and having support directly in Poudriere will be helpful to that end.

src/share/poudriere/jail.sh
kmoore134: Its not perfectly tab-aligned, would you prefer I do that or keep the formatting intact?

kmoore134:
-Gentle Nudge- 

Getting this or similar in would help with the CFT since folks won't need to manually patch:

https://lists.freebsd.org/pipermail/freebsd-pkgbase/2019-April/000396.html

Let me know if there's any additional discussion we need or want further changes to code.

src/share/poudriere/jail.sh
```diff
@@ -500,6 +511,78 @@ build_native_xtools() {
...
+	fi
+
+	# Create package for the system sources from the os/src port
+	msg "Building src: ${LOGDIR}/${JAILNAME}-buildsrc.log"
+	make -C ${PORTS_BASE}/os/src WRKDIR=${JAILMNT}/work/src BATCH=yes package >${LOGDIR}/${JAILNAME}-buildsrc.log 2>&1 || err 1 "Failed building src"
+
+	# Install the package
+	local PKGFILE="${JAILMNT}/work/src/pkg/$(make PORTSDIR=${PORTS_BASE} -C ${PORTS_BASE}/os/src -V PKGNAME).*"
+	local ABISTRING="$(make PORTSDIR=${PORTS_BASE} -C ${PORTS_BASE}/os/src -V PKG_ABISTRING)"
+	pkg -o ABI=${ABISTRING} -r ${JAILMNT} add ${PKGFILE} || err 1 "Failed installing src"
```
bdrewery: Please quote paths. `"${JAILMNT}"` (all paths in the change).

src/share/poudriere/jail.sh
```diff
@@ -500,6 +511,78 @@ build_native_xtools() {
...
+	# Cleanup the src package
+	msg "Cleaning port files"
+	make -C ${PORTS_BASE}/os/src WRKDIR=${JAILMNT}/work/src BATCH=yes clean 2>/dev/null >/dev/null || err 1 "Failed make clean"
+	make -C ${PORTS_BASE}/os/src WRKDIR=${JAILMNT}/work/src BATCH=yes distclean 2>/dev/null >/dev/null || err 1 "Failed make distclean"
+
+	for tgt in world kernel
+	do
+		# Create the package
+		msg "Building ${tgt}: ${LOGDIR}/${JAILNAME}-build${tgt}.log"
+		make -C ${PORTS_BASE}/os/build${tgt} \
```
bdrewery: How does cross-build work? I don't see any `TARGET`/`TARGET_ARCH` handling.

src/share/poudriere/common.sh
```diff
@@ -7514,6 +7517,26 @@ calculate_ospart_size() {
 	msg "OS Partiton size: ${OS_SIZE}m"
 }
 
+ingest_base_pkgs()
+{
+	# Do we have any base-port packages to ingest?
+	[ -d "${MASTERMNT}/.packages/" ] || return 0
```
bdrewery: So we build ports/packages in `jail` and then copy them in during `bulk`. Why not just build them in `bulk` like normal ports? It doesn't seem the jail is even created from these packages.

The whole build inside of `jail` concerns me as it is more unsandboxed/unjailed/shared building when we should be shrinking those cases.

bdrewery:
This was superceded by #799 which brought in support for the Makefile.inc1 support. Thank you for your contribution and I'm sorry it didn't work out. I prefer something closer to what is here rather than the current solution since it allows inline src building with bulk, but it's hard to push back against some things.

PR #687, opened 2019-05-02
Add sleep, ls and ping to miniroot

ocochard:
My miniroot /etc/rc script configure network from kenv variable, then fetch full poudriere image using fetch.
But between these two step, we need to wait for the network interface to go UP: So we need "bin/sleep".
Other user would prefer to wait for a "ping -o": So sbin/ping seems useful too.
And the last one: "bin/ls", because trying to debug miniroot without being able to use 'ls' is frustrating.

### DIFF
--- a/src/share/poudriere/image.sh
+++ b/src/share/poudriere/image.sh
@@ -90,9 +90,9 @@ mkminiroot() {
 	dirs="etc dev boot bin usr/bin libexec lib usr/lib sbin"
 	files="sbin/init etc/pwd.db etc/spwd.db"
 	files="${files} bin/sh sbin/halt sbin/fasthalt sbin/fastboot sbin/reboot"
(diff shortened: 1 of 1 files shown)


brd:
Hey Oliver,

I don't think sleep is needed, since `sh(1)` has a builtin that could be used.. Can you remove that?


Thanks,
Brad

ocochard:
Are you sure about sleep a builtin ? Because I didn't find reference to it in builtin(1), and I've did this test:
```
$ cat tutu.sh
#!/bin/sh
sleep 2
$ mv /bin/sleep /bin/sleep.bak
$ sh ./tutu.sh
./tutu.sh: sleep: not found
```

brd:
You are correct, sorry about that!

I will merge this.

PR #690, opened 2019-05-15
on mutable base do not raise securelevel errors, fixes #684

### LINKED ISSUE #684: kern.securelevel >= 1 should not prevent poudriere from working
I am successfully running builds with `kern.securelevel=2` (after commenting out the error check). Currently poudriere bails from execution skipping the check only when TMPFS_ALL being set - which imho is excessive.

I have been building bulk ports with TMPFS set to no and wrkdir, data. The only time I found schg flags breaking functionality was attempting to build jails from source - which can be circumvented by using a `-m null -M path` for custom built jails.

I would like to suggest changing the reporting from Error to Warning level or listing all actual modes that need chflags and prevent execution only if the requested feature set really needs a lower securelevel.

mulander:
Poudriere needs to use schg flags when operating on MUTABLE_BASE, however
when the schg_immutable_base() function (which properly accounts for TMPFS_ALL
NO_ZFS and actual MUTABLE_BASE setting) was introduced the securelevel check
hasn't been altered to use it, instead only checking TMPFS_ALL setting.

This prevented poudriere from running, even in situations where the requested
task did not result in attempts to use schg flags.

Signed-off-by: Adam Wolk <a.wolk@fudosecurity.com>

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -2354,8 +2354,8 @@ jail_start() {
 	if [ "${DISTFILES_CACHE}" != "no" -a ! -d "${DISTFILES_CACHE}" ]; then
 		err 1 "DISTFILES_CACHE directory does not exist. (cf.  poudriere.conf)"
 	fi
-	[ ${TMPFS_ALL} -ne 1 ] && [ $(sysctl -n kern.securelevel) -ge 1 ] && \
-	    err 1 "kern.securelevel >= 1. Poudriere requires no securelevel to be able to handle schg flags. USE_TMPFS=all can override this."
+	schg_immutable_base && [ $(sysctl -n kern.securelevel) -ge 1 ] && \
+	    err 1 "kern.securelevel >= 1. Poudriere requires no securelevel to be able to handle an immutable base (schg flags)."
 	[ "${name#*.*}" = "${name}" ] ||
 		err 1 "The jail name cannot contain a period (.). See jail(8)"
 	[ "${ptname#*.*}" = "${ptname}" ] ||


bdrewery:
This is probably right but I don't understand the old code. Why does tmpfs not care about securelevel?
And `schg_immutable_base` handles `TMPFS_ALL` slightly differently so whatever this old code was doing is changed now too.

mulander:
I'm also having trouble finding the reason why TMPFS_ALL was treated as an exception. The securelevel check has been extended in commit:

https://github.com/freebsd/poudriere/commit/01eb909ea32a1d16fcdb4afdb5f895184e92b4e1

I checked the chflags docs and found:
> Once you have set the security level to 1, write access to raw devices will be denied and special chflags(1) flags, such as schg, will be enforced.

Which made me think, perhaps tmpfs doesn't have the schg enforced. I some friends who happened to have a box at hand to run a test and it seems that with securelevel > 1 you can still set files as immutable (schg) but removing that flag (noschg) fails.

In my case, I never hit the schg failure due to always building with mutable base, as indicated with my initial #684 report - regardless what I set in TMPFS_ALL I was still able to build.

Ticket https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=226610 indicates:

> The badly named MUTABLE_BASE=schg makes the base system immutable via schg flags and is only supported for tmpfs/ufs.

Would indicate that immutable base is only supported on tmpfs/ufs. Which contradicts the hint to override the need of securelevel by using tmpfsall.

If I'm reading the current check in `schg_immutable_base` correctly then the logic is:

The filesystem is immutable when either one of is true:

- MUTABLE_BASE is set to `schg` (so it's set to `MUTABLE_BASE=no` before jail start)
- TMPFS_ALL is *not* used
- NO_ZFS is set (ZFS is *not* used)

which would indicate: immutable base system builds are only done on non zfs, non tmpfs storage when an immutable base was requested.

I guess by extension, if we knew that the base was always mutable for tmpfs we could tell the users to set TMPFS_ALL to override the need for securelevel < 1.

bdrewery:
I remembered in the middle of sleep hehe. It’s because with tmpfs I don’t have to *unchflag* because we can just *umount* when done so securelevel is respected.

PR #697, opened 2019-06-04
testport.sh: Fix make.conf not being respected

### LINKED ISSUE #545: testport -c reads /etc/make.conf
I'm not sure if this is new/expected behaviour in 3.2.0 but `poudriere testport -c` seems to be reading system's /etc/make.conf and setting default port option values based on OPTIONS_SET/OPTIONS_UNSET in it. I don't have any poudriere-specific make.conf files in /usr/local/etc/poudriere.d.

To reproduce, set OPTIONS_UNSET=DOCS in /etc/make.conf and do a `testport -c` of any port with DOCS option - both options dialog and poudriere log will show DOCS option initial value as OFF.

/etc/make.conf seems to be read only when testport is used with `-c` switch. Without -c, testport uses default option values (e.g. DOCS is ON etc) regardless of /etc/make.conf settings.

koobs:
Currently, `testport -c` does not honour make.confs such as can be set
in `poudriere.d/{*-}make.conf` files, because the `__MAKE_CONF` variable [1]
is set to /dev/null, introduced in a43fe7cb to fix #545.

Accordingly, when using testport -c, when dialog4ports is run, it presents
information that does not match what is actually used/built because the
expected make environment(s) configurations are not used.

For example, this issue was identified while testing the www/py-autobahn
port, and make config (dialog4ports) was incorrectly displaying
py36-autobahn and py36 specific OPTIONS, when testport was in fact
building py27-autobahn and py27 specific OPTIONS (correctly) via -zpy27,
which contained a DEFAULT_VERSIONS=python=2.7 override.

The issue is not limited to package names and OPTIONS, but anything that
needs to take effect or be modified in the config target.

Conversely, the options command introduced a fix to "Fix make.conf not
being respected in 2d9b712b, introducing a `setup_makeconf()` function in
common.sh that constructs a `make.conf` based on global, tree, and set
specific make.conf's, if they are set.

This change utilises the `setup_makeconf()` function in common.sh, and
applies it to testport.sh to correctly respect make.confs when -c is used.

I have tested this locally with a patch against ports-mgmt/poudriere-devel
and confirmed that it correctly finds and appends all relevant make.confs,
and presents information in dialog4ports that is correct, and matches what
is used and built during the testport.

I wasn't sure whether the following CLEANUP_HOOK from options.sh also
needed to be replicated in testport.sh:

```
CLEANUP_HOOK=options_cleanup
options_cleanup() {
    rm -f ${__MAKE_CONF}
}
```

If it does, let me know and I'll that that to the PR

[1] man make.conf

### DIFF
--- a/src/share/poudriere/testport.sh
+++ b/src/share/poudriere/testport.sh
@@ -216,6 +216,7 @@ if [ $CONFIGSTR -eq 1 ]; then
 	    make -C ${portsdir}/${ORIGIN} \
 	    ${FLAVOR:+FLAVOR=${FLAVOR}} \
 	    config
+	rm ${__MAKE_CONF}
 fi
 
 # deps_fetch_vars lookup for dependencies moved to prepare_ports()
--- a/src/share/poudriere/testport.sh
+++ b/src/share/poudriere/testport.sh
@@ -205,11 +205,10 @@ fi
 injail /usr/bin/make -C ${PORTSDIR}/${ORIGIN} maintainer ECHO_CMD=true || \
     err 1 "Port is broken"
 
-__MAKE_CONF=$(mktemp -t poudriere-make.conf)
-setup_makeconf ${__MAKE_CONF} "${JAILNAME}" "${PTNAME}" "${SETNAME}"
-
 if [ $CONFIGSTR -eq 1 ]; then
 	command -v dialog4ports >/dev/null 2>&1 || err 1 "You must have ports-mgmt/dialog4ports installed on the host to use -c."
+	__MAKE_CONF=$(mktemp -t poudriere-make.conf)
+	setup_makeconf ${__MAKE_CONF} "${JAILNAME}" "${PTNAME}" "${SETNAME}"
 	PORTSDIR=${portsdir} \
 	    __MAKE_CONF=${__MAKE_CONF} \
 	    PORT_DBDIR=${MASTERMNT}/var/db/ports \
--- a/src/share/poudriere/testport.sh
+++ b/src/share/poudriere/testport.sh
@@ -205,10 +205,13 @@ fi
 injail /usr/bin/make -C ${PORTSDIR}/${ORIGIN} maintainer ECHO_CMD=true || \
     err 1 "Port is broken"
 
+__MAKE_CONF=$(mktemp -t poudriere-make.conf)
+setup_makeconf ${__MAKE_CONF} "${JAILNAME}" "${PTNAME}" "${SETNAME}"
+
 if [ $CONFIGSTR -eq 1 ]; then
 	command -v dialog4ports >/dev/null 2>&1 || err 1 "You must have ports-mgmt/dialog4ports installed on the host to use -c."
 	PORTSDIR=${portsdir} \
-	    __MAKE_CONF=/dev/null \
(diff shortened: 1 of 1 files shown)


koobs:
If you want me to scope the directives inside the `if [ $CONFIGSTR -eq 1 ]; then` block let me know. 

I'm not familiar with the impact or need either way, but I just tested it within the block and it seems to continue to resolve the problem, but only scoped to the `make config` call

koobs:
A minor (visual) and harmless artifact, I just noticed after this is that 'Appending to make.conf' entries are printed twice, presumably once when creating jail or build environment, and once (with this PR) during make config:

```
[00:00:00] Creating the reference jail... done
[00:00:05] Mounting system devices for 12amd64-koobs-py27
[00:00:05] Mounting ports/packages/distfiles
[00:00:05] Using packages from previously failed build: /storage/poudriere/data/packages/12amd64-koobs-py27/.building
[00:00:05] Mounting packages from: /storage/poudriere/data/packages/12amd64-koobs-py27
[00:00:05] Copying /var/db/ports from: /usr/local/etc/poudriere.d/12amd64-koobs-py27-options
[00:00:05] Appending to make.conf: /usr/local/etc/poudriere.d/make.conf
[00:00:05] Appending to make.conf: /usr/local/etc/poudriere.d/py27-make.conf
/etc/resolv.conf -> /storage/poudriere/data/.m/12amd64-koobs-py27/ref/etc/resolv.conf
[00:00:05] Starting jail 12amd64-koobs-py27
[00:00:06] Ports supports: FLAVORS SELECTED_OPTIONS
[00:00:07] Appending to make.conf: /usr/local/etc/poudriere.d/make.conf
[00:00:07] Appending to make.conf: /usr/local/etc/poudriere.d/py27-make.conf
```

This may indicate a need for a minor refactor to do it (and print it) once during a testport (or underlying bulk), then applied for all later stages

src/share/poudriere/testport.sh
```diff
@@ -205,10 +205,13 @@ fi
 injail /usr/bin/make -C ${PORTSDIR}/${ORIGIN} maintainer ECHO_CMD=true || \
     err 1 "Port is broken"
 
+__MAKE_CONF=$(mktemp -t poudriere-make.conf)
+setup_makeconf ${__MAKE_CONF} "${JAILNAME}" "${PTNAME}" "${SETNAME}"
```
bdrewery: These lines are only needed inside the `if` block below. And then the `__MAKE_CONF` file should be deleted after `make` is ran.

koobs: Done

koobs:
> 
> 
> I think you missed my comment to also `rm` the file afterwards.

Ah, I interpreted the original comment as 'if its moved into the `if [ $CONFIGSTR -eq 1 ];` block then it 'will' clean it up.

Should we copy the form that exists in options.sh? That is ..

```
CLEANUP_HOOK=options_cleanup
options_cleanup() {
    rm -f ${__MAKE_CONF}
}
```

bdrewery:
Nah just `rm` directly. The `CLEANUP_HOOK` is a good idea but it may interact poorly with the embedded `bulk` handling.

bdrewery:
Thanks!

PR #706, opened 2019-08-23
Avoid having two copies of /boot in mfs iso

swills:
Given how critical space is in mfsroot, it seems to make sense to avoid
having a copy of /boot in the mfsroot since it's in the ISO anyway and
that's the one that's used for boot, not the copy in mfsroot.


bapt:
While I do agree with the approach, I disagree with the fact of removing entirely the modules. one can have a kld_list in the rc.conf so I do think we should keep the modules and nuke the kernel itself and all other files.

swills:
I agree. Perhaps it would be good to have a separate flag for this?

I've also run into cases where I need to customize the boot, such as creating /boot.config or modifying /boot/device.hints. Perhaps a separate root for the iso contents vs the mfsroot would make sense?

PR #708, opened 2019-08-28
Use sysrc to modify rc.conf

jlduran:
`touch` is required in case there is no rc.conf

### DIFF
--- a/src/share/poudriere/image.sh
+++ b/src/share/poudriere/image.sh
@@ -413,7 +413,7 @@ cap_mkdb ${WRKDIR}/world/etc/login.conf
 
 # Set hostname
 if [ -n "${HOSTNAME}" ]; then
-	echo "hostname=${HOSTNAME}" >> ${WRKDIR}/world/etc/rc.conf
+	sysrc -R ${WRKDIR}/world hostname=${HOSTNAME} > /dev/null 2>&1
 fi
 
 # Convert @flavor from package list to a unique entry of pkgname, otherwise it


jlduran:
As it has been marked for discussion:

I have pushed another revision, using sysrc's `-R` switch to set the relative root directory. Also removed the `touch`, however, if there is no _rc.conf_ (or `hostname`) defined, sysrc will complain (`awk: can't open file /etc/rc.conf`), I don't know how to address this (`> /dev/null 2>&1`)? cc/ @freebsdfrau

Thanks!

freebsdfrau:
That's a bug. I worked up the following patch to solve it:

```
--- /usr/share/bsdconfig/sysrc.subr.orig        2019-08-30 13:16:04.604403000 +0000
+++ /usr/share/bsdconfig/sysrc.subr        2019-08-30 13:14:55.058956000 +0000
@@ -560,7 +560,7 @@ f_sysrc_set()
         #
         if [ "$not_found" ]; then
                 # Add a newline if missing before appending to the file
-                awk 'BEGIN { wc = 0 } NR == 1 {
+                [ ! -e "$file" ] || awk 'BEGIN { wc = 0 } NR == 1 {
                         (cmd = "wc -l " FILENAME) | getline
                         close(cmd)
                         wc = $1
```

bdrewery:
Well there's a reason for simplicity. Would have to create the file to avoid the bug.
Anyway https://github.com/freebsd/poudriere/pull/731 will likely conflict with this change since it's changing this line of code and moving it elsewhere. Converting to `sysrc` after that PR could still be done.

freebsdfrau:
> Well there's a reason for simplicity. Would have to create the file to avoid the bug.
> Anyway #731 will likely conflict with this change since it's changing this line of code and moving it elsewhere. Converting to `sysrc` after that PR could still be done.

Merged to head 42 hours ago

https://svnweb.freebsd.org/base?view=revision&revision=355280

jlduran:
Closing this PR, as it will be addressed in #731. Also stable/11 is about to EOL in a few months.

Thank you!

PR #711, opened 2019-08-28
Reinstate save_cfg

jlduran:
After 9123631e47f9017f9a3f3de4362eea631179205a, `save_cfg` does not end in the final image, as inteded by 28604ac2492c1c61f3ac83d8812bc3d7cfdaa378.  I'm not sure if this was intentional given what is written in the comments.  Or should it be removed completely?

### DIFF
--- a/src/share/poudriere/image.sh
+++ b/src/share/poudriere/image.sh
@@ -557,6 +557,8 @@ usb)
 		rm -rf ${WRKDIR}/world/usr/local/etc
 		ln -s /etc/local ${WRKDIR}/world/usr/local/etc
 	fi
+	# Copy save_cfg to /etc
+	cp ${mnt}/usr/src/tools/tools/nanobsd/Files/root/save_cfg ${WRKDIR}/world/etc/
 	# Copy /etc and /var to /conf/base as "reference"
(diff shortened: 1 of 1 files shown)


jlduran:
I wonder why @brd used `/etc` instead of `/root` or the entire (`/usr/src/tools/tools/nanobsd/`)Files overlay dir?

bdrewery:
It doesn't look like this ever worked.

@bapt @swills @evadot @brd can one of you review this please?

jlduran:
Thank you @bdrewery!

Regarding the other Files, specifically `update*`, won't work without modifications because of GPT. But, even if adapted, the `bootme`/`bootonce` also wouldn't work because of [Bug 201788]. On top of that, https://mobile.twitter.com/f0andrey/status/1172579272873172993.

That's why I believe it's better off for now?

[Bug 201788]: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=201788

bdrewery:
Thanks for your submission. It looks like commit 2cb8ac847dd8cbce555f3d2a32c1e78f5d249ec4 (from November 2020) should fix this.

Issue #542, opened 2017-11-08
All output needs to show default FLAVOR


bdrewery:
The output for the default FLAVOR right now only shows the origin.

bdrewery:
This can then drop the `| PKGNAME` displays.

bdrewery:
The summary is done in commit 2bfcb3aab707878cadf1ff57630ae80f30b0c320 but not the dropping of `| PKGNAME`.

PR #713, opened 2019-09-04
Initial support for overlays in poudriere

bapt:
Only regular bulks tested


src/share/poudriere/testport.sh
```diff
@@ -113,6 +115,12 @@ while getopts "B:o:cniIj:J:kNp:PSvwz:" FLAG; do
 		N)
 			BUILD_REPO=0
 			;;
+		O)
+			porttree_exists ${OPTARG} ||
+			    err 2 "No such overlay ${OPTARG}"
+			OVERLAYS="${OVERLAYS} ${OPTARG}"
+			;;
+			;;
```
evadot: Redudant ';;' ?

src/share/poudriere/common.sh
```diff
@@ -6495,6 +6526,7 @@ compute_deps_pkg() {
 				dep_pkgname="${key%%[><=]*}"
 				dpath="${d#*:}"
 				case "${dpath}" in
+					# XXXX
```
evadot: what's this ?

src/share/poudriere/testport.sh
bapt: done

src/share/poudriere/common.sh
bapt: done

bdrewery:
Ports piece at https://reviews.freebsd.org/D21468

bdrewery:
This doesn't cover `MOVED`

src/share/poudriere/common.sh
```diff
@@ -6646,6 +6689,37 @@ listed_ports() {
...
+	local ptdir=$1
+	# skip overlays with no categories listed
+	[ -f ${ptdir}/Makefile ] || return
+	for cat in $(awk -F= '$1 ~ /^[[:space:]]*SUBDIR[[:space:]]*\+/ {gsub(/[[:space:]]/, "", $2); print $2}' ${ptdir}/Makefile); do
+		# skip overlays with no ports hooked to the build
+		[ -f ${ptdir}/${cat}/Makefile ] || continue
+		awk -F= -v cat=${cat} '$1 ~ /^[[:space:]]*SUBDIR[[:space:]]*\+/ {gsub(/[[:space:]]/, "", $2); print cat"/"$2}' ${ptdir}/${cat}/Makefile
+	done | while mapfile_read_loop_redir origin; do
+		if ! [ -d "${ptdir}/${origin}" ]; then
+			msg_warn "Nonexistent origin listed in category Makefiles: ${COLOR_PORT}${origin}${COLOR_RESET} (skipping)"
```
bdrewery: Anywhere there's a warning or assertion for a problem in a port now needs to display which `OVERLAY` it's in.

bapt: done

bapt:
> The wording is probably confusing if people don't know what an overlay is. What about something like `Specify path to overlay onto the ports tree`

done

bapt:
> Items to be resolved before merge to master
> 
>     * `MOVED` support

You want to support a MOVED file per overlay? if yes this was not initially planned, but yes we can do it.
> 
>     * Needs tests added
>

Oh I haven't seen there was a testsuite now. I will try to add something in there.

>     * manpage updates
> 

Sure.

bdrewery:
I'm willing to write the tests portion. It'll be a little tricky since the ports tree checked in is partial, old, and hacked up. The tests are quite funky in how they work.

`MOVED` is something that makes sense. It's handled by the tool not the framework. I've wanted a `MOVED` merging overlay for quite a while now. But yes it can wait until later.

bdrewery:
> Port directory / Origin messages need to show which OVERLAY it's in

This piece will be annoying to maintain even though I think it's needed. I'm not sure how it will even work without adding the chosen `OVERLAY` to the "originspec" thing (origin@dep_args@flavor) handled in `originspec_encode` and `originspec_decode`.
It may just have to not be in the first release.

bdrewery:
Sorry. I'll look at this serious tomorrow and try to get it in.

src/share/poudriere/common.sh
```diff
@@ -1466,6 +1483,7 @@ common_mtree() {
 	./var/db/etcupdate
 	./var/db/ports
 	./wrkdirs
+	./overlays
```
bdrewery: This is a sorted list apparently

bdrewery:
Thanks for working on this. I have to wonder how usable unionfs would be here since the ports tree is *read only*. It would seemingly be trivial if it doesn't have to worry about modifications at any layer. Seems like a pretty straightforward fuse module too.

Having reviewed this more deeply now I see how trivial it is overall. Just need to fix absolute path references to use the new portdir lookup function.

I know there's some `OVERLAY` support in the ports tree review but if unionfs works here then the total patch should be a few lines in Poudriere to pass `bulk -O` through to `do_portbuild_mounts`. I'm going to try that a bit first before merging this.

bdrewery:
```
~/git/poudriere # mount -t unionfs -o ro /root/svn/ports /tmp/ports
```
```
KDB: stack backtrace:
db_trace_self_wrapper() at db_trace_self_wrapper+0x2b/frame 0xfffffe24a96fe460
assert_vop_elocked() at assert_vop_elocked+0x72/frame 0xfffffe24a96fe490
insmntque1() at insmntque1+0x3e/frame 0xfffffe24a96fe4d0
unionfs_nodeget() at unionfs_nodeget+0x2b1/frame 0xfffffe24a96fe560
unionfs_domount() at unionfs_domount+0x5ad/frame 0xfffffe24a96fe7b0
vfs_domount() at vfs_domount+0x7fe/frame 0xfffffe24a96fe9e0
vfs_donmount() at vfs_donmount+0x88d/frame 0xfffffe24a96fea80
sys_nmount() at sys_nmount+0x69/frame 0xfffffe24a96feac0
amd64_syscall() at amd64_syscall+0x2b9/frame 0xfffffe24a96febf0
fast_syscall_common() at fast_syscall_common+0x101/frame 0xfffffe24a96febf0
--- syscall (378, FreeBSD ELF64, sys_nmount), rip = 0x8002d29da, rsp = 0x7fffffffdbf8, rbp = 0x7fffffffe580 ---
vnode 0xfffff8100688e000: tag unionfs, type VNON
    usecount 1, writecount 0, refcount 1
    flags ()
    lock type unionfs: UNLOCKED
#0 0xffffffff80b9f41a at lockmgr_lock_fast_path+0x20a
#1 0xffffffff810e02a0 at VOP_LOCK1_APV+0x70
#2 0xffffffff80ca92f8 at vputx+0x278
#3 0xffffffff80cb3f83 at kern_funlinkat+0x313
#4 0xffffffff80cb3c68 at sys_unlink+0x28
#5 0xffffffff828ba5c6 at filemon_wrapper_unlink+0x16
#6 0xffffffff81042979 at amd64_syscall+0x2b9
#7 0xffffffff81019ed0 at fast_syscall_common+0x101
insmntque: non-locked vp: 0xfffff8100688e000 is not exclusive locked but should be
KDB: enter: lock violation
[ thread pid 73134 tid 101450 ]
Stopped at      kdb_enter+0x3b: movq    $0,kdb_why
```
RIP

How do the live cds even work anymore then?

bdrewery:
I'm merging this outside of github. Will push it after fixing up some missed `_lookup_portdir` and other minor nits.

bdrewery:
Oh there's a branch. Will just fixup there instead.

bdrewery:
There's a ton of `${PORTSDIR}/foo` and `${portsdir}/foo` references still. I'm fixing them up and then will merge this.

bdrewery:
Current issues:

- [ ] make.conf, poudriere.conf, options, blacklist support for each OVERLAY
- [ ] `clean_restricted()` is only running at the parent level of the main tree. Should change to use `listed_ports` and parallelize it for all OVERLAYS
- [ ] `listed_ports()` will do the wrong thing in `-a` mode because it can list some ports twice but that's asserted to be disallowed later.
- [ ] `_lookup_portdir` areas near `msg*` need to better show the OVERLAY. This should be done with #542 which needs to add better functions to handle printing package/origins.

PR #728, opened 2019-10-29
Set DEVELOPER in patch phase

swills:
Allow potential QA checks to be run in patch phase


bdrewery:
For my reference what's an example test would this enable that `check-sanity` wasn't picking up?

bdrewery:
Looks like https://reviews.freebsd.org/D22174

swills:
Yes, please merge it?

PR #731, opened 2019-11-28
Add support for zfs send stream as an output format for system images

allanjude:
Includes support for sending an entire pool, or just a boot environment (for upgrades)
Adds the -R flag to allow user to specify ZFS send features

It uses the existing pre/post scripting to control the finer details

Add support for specifying multiple zfssend targets simultaneously.
Document changes in man page and command help output.
poudriere image is a bit too eager setting hostname
	Old behaviour:
	- poudriere image would always set hostname
	New behaviour:
	- If hostname is specified with the -h command line option: Always set hostname
	- Otherwise, and only if image doesn't already specify a hostname (see: option -c overlaydir), set default hostname.

Sponsored by: Modirum MDPay
Sponsored by: Klara Systems


src/share/poudriere/common.sh
```diff
@@ -7661,6 +7661,35 @@ calculate_ospart_size() {
 	msg "OS Partiton size: ${OS_SIZE}m"
 }
 
+
+# Stolen from bsdconfig
+# str_replaceall $string $find $replace [$var_to_set]
```
bdrewery: We have a function for this already:
```
_gsub "$string" "$match" "$replace"
# $_gsub contains result
# or use result=$(gsub ...)
```
I just pushed this new version to match the API you used. So you should just be able to change the function call name to `_gsub`.
```
_gsub "$string" "$match" "$replace" "$output"
```

src/share/poudriere/image.sh
```diff
@@ -130,10 +135,83 @@ mkminiroot() {
...
+		# Create the dataset with desired options
+		zfs create ${options:+-o $options} "$zroot$dataset" ||
+		    err 1 "Dataset creation failed"
+	done
+}
+
+zfssend_writereplicationstream() {
+	# Arguments:
+	# $1: snapshot to recursively replicate
+	# $2: Image name to write replication stream to
```
bdrewery:
```
[ $# -eq 2 ] || eargs zfssend_writereplicationstream snapshot_ftom image_to
```
Causes an error if called wrong and is kinda self-documenting. Keep your comments if you want.

src/share/poudriere/image.sh
```diff
@@ -411,9 +521,16 @@ if [ -f "${WRKDIR}/world/etc/login.conf.orig" ]; then
...
 if [ -n "${HOSTNAME}" ]; then
-	echo "hostname=${HOSTNAME}" >> ${WRKDIR}/world/etc/rc.conf
+	setimagehostname "${HOSTNAME}"	
+else
+	# If hostname wasn't set on command line, only set
+	# default hostname if hostname isn't specified by
+	# overlay file strucutre.
+	if ! sysrc -n -R "${WRKDIR}/world" hostname; then
+		setimagehostname "poudriere-image"
+	fi
```
bdrewery: I think this could be simplified like this to only call `setimagehostname` once.
```
: ${HOSTNAME:=$(sysrc -n -q -R "${WRKDIR}/world" hostname || echo "poudriere-image")}
```
It would use the `-h hostname` value by default, then look at sysrc for the hostname, then fallback to the default.

src/share/poudriere/image.sh
```diff
@@ -130,10 +135,83 @@ mkminiroot() {
...
+		/var		mountpoint=/var canmount=off
+		/var/audit	exec=off setuid=off
+		/var/crash	exec=off setuid=off
+		/var/log	exec=off setuid=off
+		/var/mail	atime=on
+		/var/tmp	setuid=off
+	"}
+
+	msg "Creating ZFS Datasets"
+	echo "$ZFS_DATASETS" | while read dataset options; do
```
bdrewery: I suspect there's an elegant `awk` or `sed` version of this. It's fine as is though.

src/share/poudriere/image.sh
```diff
@@ -130,10 +135,83 @@ mkminiroot() {
...
+		# Replace tabs with spaces
+		str_replaceall "$options" "	" " " options
+		# Reduce contiguous runs of space to one single space
+		oldoptions=
+		while [ "$oldoptions" != "$options" ]; do
+			oldoptions="$options"
+			str_replaceall "$options" "  " " " options
+		done
+		# Replace both commas and spaces with ` -o '
+		str_replaceall "$options" "[ ,]" " -o " options
```
bdrewery: What about paths that have spaces? Should the spec have quoting support?

allanjude:
I've managed to get these 2 pull requests (this and #761) mixed together, but, I think it is relatively clean now

allanjude:
I managed to sepatate things back apart, but I still need to fixup the man page.

src/share/poudriere/image_zfs.sh
```diff
@@ -0,0 +1,182 @@
...
+		esac
+		;;
+	esac
+
+	zpool export ${zroot}
+	zroot=
+	/sbin/mdconfig -d -u ${md#md}
+	md=
+
+	## When generating a disk image instead, we need to export the pool first.
```
evadot, line 150: This comment should be 5 lines up.

PR #732, opened 2019-11-28
un-bork poudriere image

dch:
- `recursecopylib` fails to locate dependencies for miniroot
- `make -C` invocation is probably only needed for from-source builds

I am not certain that either of these patches are correct but they
definitely work for me here. I'm happy to amend as required.


src/share/poudriere/image.sh
```diff
@@ -402,7 +402,7 @@ touch ${WRKDIR}/src.conf
 [ ! -f ${POUDRIERED}/${JAILNAME}-src.conf ] || cat ${POUDRIERED}/${JAILNAME}-src.conf >> ${WRKDIR}/src.conf
 [ ! -f ${POUDRIERED}/image-${JAILNAME}-src.conf ] || cat ${POUDRIERED}/image-${JAILNAME}-src.conf >> ${WRKDIR}/src.conf
 [ ! -f ${POUDRIERED}/image-${JAILNAME}-${SETNAME}-src.conf ] || cat ${POUDRIERED}/image-${JAILNAME}-${SETNAME}-src.conf >> ${WRKDIR}/src.conf
-make -C ${mnt}/usr/src DESTDIR=${WRKDIR}/world BATCH_DELETE_OLD_FILES=yes SRCCONF=${WRKDIR}/src.conf delete-old delete-old-libs
+make -C ${mnt}/usr/src DESTDIR=${WRKDIR}/world BATCH_DELETE_OLD_FILES=yes SRCCONF=${WRKDIR}/src.conf delete-old delete-old-libs ||
```
bdrewery: `make` or what? Need something after `||` like `|| :`.
```
# sh
# false ||
> ^Dsh: Syntax error: end of file unexpected
```
I think it's better to check for the Makefile instead of ignoring the error.
```
[ -f "${mnt}/usr/src/Makefile" ] && make -C ...
```

src/share/poudriere/image.sh
```diff
@@ -81,7 +81,7 @@ recursecopylib() {
 		fi
 		;;
 	esac
-	for i in $( (readelf -d ${mroot}/$path 2>/dev/null || :) | awk '$2 == "NEEDED" { gsub(/\[/,"", $NF ); gsub(/\]/,"",$NF) ; print $NF }'); do
+	for i in $( (readelf -d ${mroot}/$path 2>/dev/null || :) | sed -E -e '/NEEDED/!d' -e 's/.+\[//g' -e 's/\]$//g'); do
```
bdrewery: What do you mean it no longer works? What happens for you? Did you replace `awk` on your system?
```
# readelf -d /bin/sh | awk '$2 == "NEEDED" { gsub(/\[/,"", $NF ); gsub(/\]/,"",$NF) ; print $NF }'
libedit.so.7
libc.so.7
```

dch: no, this is a boring 13.0-CURRENT with almost no [tweaks](https://hackmd.io/@dch/r1X8LNVaB#conf-tweaks). that awk line produces no output whatsoever. On my 12.1R, it's obviously all good. I have no idea why this would be different.

```
# sh
# uname -a
FreeBSD wintermute 13.0-CURRENT FreeBSD 13.0-CURRENT r355322+d838b917e79b-c264794(master) GENERIC  amd64
# readelf -d /bin/sh | sed -E -e '/NEEDED/!d' -e 's/.+\[//g' -e 's/\]$//g'
libedit.so.7
libc.so.7
# readelf -d /bin/sh | awk '$2 == "NEEDED" { gsub(/\[/,"", $NF ); gsub(/\]/,"",$NF) ; print $NF }'
# which awk
/usr/bin/awk
#  readelf -d /bin/sh

Dynamic section at offset 0x27030 contains 25 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libedit.so.7]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.7]
 0x000000000000001e (FLAGS)              STATIC_TLS
 0x0000000000000015 (DEBUG)              0x0
 0x0000000000000007 (RELA)               0x201a30
 0x0000000000000008 (RELASZ)             144 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x0000000000000017 (JMPREL)             0x201ac0
 0x0000000000000002 (PLTRELSZ)           3024 (bytes)
 0x0000000000000003 (PLTGOT)             0x2280e8
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000006 (SYMTAB)             0x200308
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000005 (STRTAB)             0x20158c
 0x000000000000000a (STRSZ)              1186 (bytes)
 0x000000006ffffef5 (GNU_HASH)           0x201100
 0x0000000000000004 (HASH)               0x20114c
 0x0000000000000019 (INIT_ARRAY)         0x227028
 0x000000000000001b (INIT_ARRAYSZ)       8 (bytes)
 0x000000000000000c (INIT)               0x22654c
 0x000000000000000d (FINI)               0x22655c
 0x000000006ffffff0 (VERSYM)             0x200fb0
 0x000000006ffffffe (VERNEED)            0x2010c0
 0x000000006fffffff (VERNEEDNUM)         1
 0x0000000000000000 (NULL)               0x0
```

src/share/poudriere/image.sh
dch: absolutely better. fixed & pushed.

src/share/poudriere/image.sh
dch: ok kevans91 helped me track this down on irc. it's a leaky user PATH that gets in via sudo and lets in /usr/local/bin/readelf, which picks up a GNU one from binutils, via gambit-c.

kevans91:
Replacing `$2 == "NEEDED"` with `match($2, "\\(?NEEDED\\)?")` would be an easy win for compatibility with llvm/binutils readelf in addition to elftoolchain.

src/share/poudriere/image.sh
```diff
@@ -402,7 +402,7 @@ touch ${WRKDIR}/src.conf
 [ ! -f ${POUDRIERED}/${JAILNAME}-src.conf ] || cat ${POUDRIERED}/${JAILNAME}-src.conf >> ${WRKDIR}/src.conf
 [ ! -f ${POUDRIERED}/image-${JAILNAME}-src.conf ] || cat ${POUDRIERED}/image-${JAILNAME}-src.conf >> ${WRKDIR}/src.conf
 [ ! -f ${POUDRIERED}/image-${JAILNAME}-${SETNAME}-src.conf ] || cat ${POUDRIERED}/image-${JAILNAME}-${SETNAME}-src.conf >> ${WRKDIR}/src.conf
-make -C ${mnt}/usr/src DESTDIR=${WRKDIR}/world BATCH_DELETE_OLD_FILES=yes SRCCONF=${WRKDIR}/src.conf delete-old delete-old-libs
+[ -f "${mnt}/usr/src/Makefile" ] && make -C ${mnt}/usr/src DESTDIR=${WRKDIR}/world BATCH_DELETE_OLD_FILES=yes SRCCONF=${WRKDIR}/src.conf delete-old delete-old-libs
```
bdrewery, line 405: I think more thought is needed here.

1. Why are we setting up a src.conf if it won't be used?
2. What about the /usr/src reference at `OSVERSION=$(awk -F '"' '/REVISION=/ { print $2 }' ${mnt}/usr/src/sys/conf/newvers.sh | cut -d '.' -f 1)`
3. and `install ${mnt}/usr/src/tools/tools/nanobsd/Files/root/save_cfg ${WRKDIR}/world/conf/base/etc/`

bdrewery (changes requested):

PR #744, opened 2020-02-10
Add documentation about OPTIONS_(UNSET|SET)

markand:
I propose a friendly documentation about default options which are only documented in /usr/ports/Mk/bsd.port.mk which users may not have since poudriere handles ports by itself as a "black-box"


bdrewery:
I think this belongs in the handbook and a link to the handbook section on options here instead.

PR #751, opened 2020-04-13
Add support for hashed mode to Poudriere

allanjude:
Requires the related patches to pkg and the ports tree

### DIFF
--- a/src/man/poudriere-bulk.8
+++ b/src/man/poudriere-bulk.8
@@ -28,7 +28,7 @@
 .\"
 .\" Note: The date here should be updated whenever a non-trivial
 .\" change is made to the manual page.
-.Dd July 5, 2022
+.Dd September 26, 2022
 .Dt POUDRIERE-BULK 8
 .Os
 .Sh NAME
@@ -38,7 +38,7 @@
 .Nm
 .Fl a
 .Fl j Ar name
-.Op Fl CcFIikNnRrSTtvw
+.Op Fl CcFHIikNnRrSTtvw
 .Op Fl B Ar name
 .Op Fl b Ar branch
 .Op Fl J Ar maxjobs Ns Op Cm \&: Ns Ar prebuildmaxjobs
@@ -48,7 +48,7 @@
 .Nm
 .Fl f Ar file Op Fl f Ar file2 Ar ...
 .Fl j Ar name
-.Op Fl CcFIikNnRrSTtvw
+.Op Fl CcFHIikNnRrSTtvw
 .Op Fl B Ar name
 .Op Fl b Ar branch
 .Op Fl J Ar maxjobs Ns Op Cm \&: Ns Ar prebuildmaxjobs
@@ -57,7 +57,7 @@
 .Op Fl z Ar set
 .Nm
 .Fl j Ar name
-.Op Fl CcFIikNnRrSTtvw
+.Op Fl CcFHIikNnRrSTtvw
 .Op Fl B Ar name
 .Op Fl b Ar branch
 .Op Fl J Ar maxjobs Ns Op Cm \&: Ns Ar prebuildmaxjobs
@@ -232,6 +232,8 @@ Fetch only from the original
 Skip
 .Fx
 mirrors.
+.It Fl H
+Create a repository where the package filenames contain the short hash of the contents.
 .It Fl I
 Advanced interactive mode.
 .Pp
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -6473,6 +6473,10 @@ delete_old_pkg() {
 		if [ -L "${pkg}" ]; then
 			is_sym=1
 		fi
+		if [ -d "${pkg}" ] && [ "${pkgfile}" = "Hashed" ]; then
+			msg_debug "Ignoring directory"
+			return 0;
+		fi
 		if [ "${is_sym}" -eq 1 ] && [ ! -e "${pkg}" ]; then
 			msg "Deleting ${COLOR_PORT}${pkgfile}${COLOR_RESET}: dead symlink"
 			delete_pkg "${pkg}"
@@ -9468,11 +9472,15 @@ clean_restricted() {
 }
 
 build_repo() {
-	local origin pkg_repo_list_files
+	local origin pkg_repo_list_files hashcmd
 
 	msg "Creating pkg repository"
 	if [ ${DRY_RUN} -eq 1 ]; then
 		return 0
+	fi
+	if [ ${PKG_HASH} != "no" ]; then
+		hashcmd="--hash --symlink"
+		PKG_REPO_FLAGS="${PKG_REPO_FLAGS:+${PKG_REPO_FLAGS} }$hashcmd"
 	fi
 	bset status "pkgrepo:"
 	ensure_pkg_installed force_extract || \
@@ -9493,12 +9501,20 @@ build_repo() {
 		install -m 0400 "${PKG_REPO_META_FILE}" \
 		    "${MASTERMNT:?}/tmp/pkgmeta"
 	fi
+
+	# Remount rw
+	# mount_nullfs does not support mount -u
+	umount ${UMOUNT_NONBUSY} ${MASTERMNT}/packages || \
+	    umount -f ${MASTERMNT}/packages
+	mount_packages
+
 	mkdir -p ${MASTERMNT}/tmp/packages
 	if [ -n "${PKG_REPO_SIGNING_KEY}" ]; then
 		msg "Signing repository with key: ${PKG_REPO_SIGNING_KEY}"
 		install -m 0400 "${PKG_REPO_SIGNING_KEY}" \
 			"${MASTERMNT:?}/tmp/repo.key"
 		injail ${PKG_BIN:?} repo \
+			${PKG_REPO_FLAGS} \
 			${pkg_repo_list_files:+"${pkg_repo_list_files}"} \
 			-o /tmp/packages \
 			${PKG_META} \
@@ -9515,6 +9531,7 @@ build_repo() {
 		# using SSH with DNSSEC as older hosts don't support
 		# it.
 		${MASTERMNT:?}${PKG_BIN:?} repo \
+		    ${PKG_REPO_FLAGS} \
 		    ${pkg_repo_list_files:+"${pkg_repo_list_files}"} \
 		    -o "${MASTERMNT:?}/tmp/packages" ${PKG_META_MASTERMNT} \
 		    "${MASTERMNT:?}/packages" \
@@ -9527,6 +9544,7 @@ build_repo() {
 			;;
 		esac
 		JNETNAME="n" injail ${PKG_BIN:?} repo \
+		    ${PKG_REPO_FLAGS} \
 		    ${pkg_repo_list_files:+"${pkg_repo_list_files}"} \
 		    -o /tmp/packages ${PKG_META} /packages \
 		    ${SIGNING_COMMAND:+signing_command: ${SIGNING_COMMAND}} ||
@@ -9542,6 +9560,11 @@ build_repo() {
 			sign_pkg pubkey "${PACKAGES:?}/Latest/pkg.${PKG_EXT}"
 		fi
 	fi
+
+	# Remount ro
+	umount ${UMOUNT_NONBUSY} ${MASTERMNT}/packages || \
+	    umount -f ${MASTERMNT}/packages
+	mount_packages -o ro
 }
 
 calculate_size_in_mb() {
@@ -10109,6 +10132,7 @@ esac
 : ${FLAVOR_DEFAULT_ALL:=no}
 : ${NULLFS_PATHS:="/rescue /usr/share /usr/tests /usr/lib32"}
 : ${PACKAGE_FETCH_URL:="pkg+http://pkg.FreeBSD.org/\${ABI}"}
+: ${PKG_HASH:=no}
 
 : ${POUDRIERE_TMPDIR:=$(command mktemp -dt poudriere)}
 : ${SHASH_VAR_PATH_DEFAULT:=${POUDRIERE_TMPDIR}}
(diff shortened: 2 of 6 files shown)


src/share/poudriere/common.sh
```diff
@@ -5280,7 +5283,7 @@ delete_old_pkg() {
 	# Check if the compiled options match the current options from make.conf and /var/db/ports
 	if [ "${CHECK_CHANGED_OPTIONS}" != "no" ]; then
 		if have_ports_feature SELECTED_OPTIONS; then
-			shash_remove pkgname-options "${new_pkgname}" \
+			shash_get pkgname-options "${new_pkgname}" \
```
bdrewery, line 5286: Is this intentional? The `shash_remove` fetches and removes the cache item. The `shash_get` only fetches. This cache is not persistent across builds. It's just for IPC.

bdrewery:
Can you also give more context for what this is?

src/share/poudriere/common.sh
bdrewery: Sorry that it's not documented.

allanjude: Yes, because we now have 2 copies of every package in the repo directory, the second one was not finding the cached options, and it was triggering rebuilds for 'changed options'. This change means that the 2nd copy of the pkg will still find the options in the cache.

allanjude:
> 
> 
> Can you also give more context for what this is?

When combined with https://github.com/freebsd/pkg/pull/1829

This creates a pkg repo that looks like this:
```
#ls -al /usr/local/poudriere/data/packages/121amd64-default/All/
total 12021
drwxr-xr-x  2 root    wheel       16 Apr 13 20:06 ./
drwxr-xr-x  4 root    wheel        9 Apr 13 20:06 ../
-rw-r--r--  1 nobody  wheel   161820 Apr 13 19:27 gettext-runtime-0.20.1+ec3887d3a2.txz
lrwxr-xr-x  1 nobody  wheel       37 Apr 13 19:27 gettext-runtime-0.20.1.txz@ -> gettext-runtime-0.20.1+ec3887d3a2.txz
-rw-r--r--  1 nobody  wheel  2527876 Apr 13 19:29 gettext-tools-0.20.1_1+7101c9ffe5.txz
lrwxr-xr-x  1 nobody  wheel       37 Apr 13 19:29 gettext-tools-0.20.1_1.txz@ -> gettext-tools-0.20.1_1+7101c9ffe5.txz
-rw-r--r--  1 nobody  wheel     5828 Apr 13 19:19 indexinfo-0.3.1+1cd9c1a735.txz
lrwxr-xr-x  1 nobody  wheel       30 Apr 13 19:19 indexinfo-0.3.1.txz@ -> indexinfo-0.3.1+1cd9c1a735.txz
-rw-r--r--  1 nobody  wheel   387088 Apr 13 19:27 libtextstyle-0.20.1+6c117ad74e.txz
lrwxr-xr-x  1 nobody  wheel       34 Apr 13 19:27 libtextstyle-0.20.1.txz@ -> libtextstyle-0.20.1+6c117ad74e.txz
-rw-r--r--  1 nobody  wheel   236640 Apr 13 20:04 nano-4.8+74c3c14712.txz
lrwxr-xr-x  1 nobody  wheel       23 Apr 13 20:04 nano-4.8.txz@ -> nano-4.8+74c3c14712.txz
-rw-r--r--  1 nobody  wheel  8787688 Apr 13 20:06 pkg-1.13.99.7.l+46fd66c8a7.txz
lrwxr-xr-x  1 nobody  wheel       30 Apr 13 20:06 pkg-1.13.99.7.l.txz@ -> pkg-1.13.99.7.l+46fd66c8a7.txz
-rw-r--r--  1 nobody  wheel    35452 Apr 13 19:54 zxfer-1.1.7+2e39bad872.txz
lrwxr-xr-x  1 nobody  wheel       26 Apr 13 19:54 zxfer-1.1.7.txz@ -> zxfer-1.1.7+2e39bad872.txz
```

So when you install a package, it fetches the file with the hash in the URL:
```
# pkg install zxfer
Updating Test repository catalogue...
Test repository is up to date.
All repositories are up to date.
The following 1 package(s) will be affected (of 0 checked):

New packages to be INSTALLED:
        zxfer: 1.1.7

Number of packages to be installed: 1

35 KiB to be downloaded.

Proceed with this action? [y/N]: y
[1/1] Fetching zxfer-1.1.7+2e39bad872.txz: 100%   35 KiB  35.5kB/s    00:01
Checking integrity... done (0 conflicting)
[1/1] Installing zxfer-1.1.7...
[1/1] Extracting zxfer-1.1.7: 100%
```


This will allow package repositories to be served from CDNs and web caches, since the unique hash in the filename will avoid the need for cache invalidation on the actual package files, and a short lifetime on the pkg meta files is all that would be required.

src/share/poudriere/common.sh
bdrewery: Well doubling the startup time isn't wanted and this will be hard to maintain. Perhaps symlinks should be trimmed out (by changing the glob *.PKG_EXT to a `$(cd dir && find . ...)` kind of thing) or % files should be skipped in `delete_old_pkg`.

Do we even need the symlinks?

src/share/poudriere/common.sh
```diff
@@ -5090,6 +5090,9 @@ delete_old_pkg() {
 	fi
 
 	v="${pkgname##*-}"
+	# Strip the hash from the filename to leave just the version
+	v="${v%+*}"
```
bdrewery, line 5094:
```
% pkg rquery %n-%v|grep -F +
afl++-2.62c
arpack++-2.3.0
arts++-1.1.a13_2,1
blitz++-1.0.2
bonnie++-1.98
bpl+-1.0_1
c++-gtk-utils-2.2.18
clisp-2.49.93+_1
crf++-0.54_1
cvs+ipv6-1.20120905_3
dbus-c++-0.9.1_5
dia2code+-1.0.0_7
dvd+rw-tools-7.1_3
exim-sa-exim-4.93.0.4+4.2.1_3
fc++-1.5
flvtool++-1.2.1_14
gmm++-5.3
golded+-1.1.5.a20170303
ja-timidity++-slang-2.15.0
lapack++-2.5.4_9
libcdio-paranoia-10.2+2.0.1
libodbc++-0.2.5_5
libsigc++-2.10.2
libsigc++12-1.2.7_2
libstdc++_stldoc_4.2.2-20071101_1
libxml++-2.40.1,1
linux-c7-libsigc++20-2.10.0
lucene++-3.0.7_14
make++-2.0
memtest86+-5.01_3
mpfrc++-3.6.2
mysql-connector-c++-1.1.13_1
ncbi-blast+-2.10.0
nurbs++-3.0.11_9
ocaml-camlp4-4.05+2
p5-Text-Format+NWrap-0.52_1
p5-Text-Tabs+Wrap-2013.0523_1
pcre++-0.9.5_5
png++-0.2.10
quickcheck++-0.0.3
s2tc-1.0+20151228_1
sdts++-1.5.1_19
smm++-6.1.1_1
timidity++-2.15.0
timidity++-emacs-2.15.0
timidity++-gtk-2.15.0
timidity++-motif-2.15.0
timidity++-slang-2.15.0
timidity++-xaw-2.15.0
timidity++-xskin-2.15.0
tintin++-2.02.02
tolua++-1.0.93_3
vera++-1.3.0_14
voro++-0.4.6_1
weblint++-1.15_3
wmweather+-2.17_1
xeyes+-2.02_2
xgalaga++-0.9
xlife++-2.0.1_5
yaz++-1.6.1_1
zh-tintin++-2.02.02
```

src/share/poudriere/common.sh
allanjude: I wanted to not have the symlink, but that was causing a lot of other fallout.

Since now the filename is unpredictable, things like `clean_build_queue()` can only look for the symlink, since they cannot guess the hashed name of the file.

src/share/poudriere/common.sh
allanjude: The line above has already tripped off everything before the last -, so it is just %v, but there are still a few hits there:
2.49.93+_1
4.93.0.4+4.2.1_3
10.2+2.0.1
4.05+2
1.0+20151228_1

src/share/poudriere/common.sh
allanjude: The need for this has been overcome by moving the hashed files into the .hashed directory, to avoid duplicate detection

src/share/poudriere/common.sh
allanjude: As agreed on IRC, switched to using ~ instead of % or +

allanjude:
@bapt with https://github.com/freebsd/pkg/commit/36dfb4895b0b307fe3321f6a4478521080a6e9b7 merged into pkg, I've refreshed this patch to add a -H flag to poudriere bulk, which builds a repo using the hashed mode.

It currently implies --symlink as well, because poudriere doesn't find the already build packages during an incremental without it yet.

allanjude:
The change committed to pkg is different than the original proposal (create hashed filenames during `pkg create`). The version that was merged to pkg is for `pkg repo` which does all the work in one step at the end, and requires a lot fewer changes to poudriere that way.

darkfiberiru:
@allanjude Can we also have a poudriere.conf knob.  Below is from my attempt I will write and test an additional commit to work with yours to do that.

> # Have pkg create hashed versions of the pkg filenames with symlinks to
> # original pkg names. The packagesite.yaml file will point to the hashed version
> # of these files. By using hashed pkg filenames, this allows users to lazily
> # synchronise packages without conflicting with the current packages,
> # for example using rsync or CDNs.  Once the packages are synced the much
> # smaller meta files can then be synced. Allowing a near atomic update of repo.
> # On caching cdn this means a need to purge 2-5 files instead of all pkgs that
> # have been updated.
> #PKG_HASH=no

darkfiberiru:
@allanjude As discussed oob I will try to get a patch version that includes -H flag or poudriere.conf options and generate a new pr/reopen #786

igalic:
@darkfiberiru do you still have plans to pick this up again?
it seems Allan is perpetually busy with something else

allanjude:
One thing I noticed, with the new default config, the pkgs get owned by 'nobody', but the symlink's to the hashed versions are owned by root.
Is this a case of `pkg repo` should be run as nobody, or that we just need to do a chown after pkg repo?

I notice packagesite etc are not owned by nobody.

allanjude:
I have refreshed this and solved the issues that have been reported (Poudriere was trying to delete the Hashed directory as it thought it was an orphaned package)

bdrewery:
Looking at this today.

src/share/poudriere/common.sh
```diff
@@ -9493,12 +9501,20 @@ build_repo() {
 		install -m 0400 "${PKG_REPO_META_FILE}" \
 		    "${MASTERMNT:?}/tmp/pkgmeta"
 	fi
+
+	# Remount rw
+	# mount_nullfs does not support mount -u
+	umount ${UMOUNT_NONBUSY} ${MASTERMNT}/packages || \
```
bdrewery, line 9507: See 43cca931c95fbdeccfa9d93a82ee0dd8110055b2
s/${UMOUNT_NONBUSY}/-n/

src/share/poudriere/common.sh
```diff
@@ -9542,6 +9560,11 @@ build_repo() {
 			sign_pkg pubkey "${PACKAGES:?}/Latest/pkg.${PKG_EXT}"
 		fi
 	fi
+
+	# Remount ro
+	umount ${UMOUNT_NONBUSY} ${MASTERMNT}/packages || \
```
bdrewery, line 9565: UMOUNT_NONBUSY to -n here too. Or better `remount_packages -o ro`.

src/share/poudriere/include/pkg.sh
```diff
@@ -417,6 +421,8 @@ delete_pkg_xargs() {
 	# Delete the package and the depsfile since this package is being deleted,
 	# which will force it to be recreated
 	{
+		# If ${pkg} is a symlink, delete the target as well
+		[ -L "${pkg}" ] && echo $(realpath "${pkg}")
```
bdrewery, line 425: Please avoid `&&` as if it ends up being the last in a statement the statement can return non-zero with `set -e`. So we avoid it as a pattern because it sneaks in eventually.
```
# sh -c 'set -o pipefail; set -e; dolink() { [ -L / ] && true; }; dolink; echo done'; echo $?
1

File: src/man/poudriere-bulk.8 [hunk 6]
```diff
@@ -28,7 +28,7 @@
 .\"
 .\" Note: The date here should be updated whenever a non-trivial
 .\" change is made to the manual page.
-.Dd July 5, 2022
+.Dd September 26, 2022
```
bdrewery, line 31: Update to more recent given the PR idle time.

src/share/poudriere/common.sh
```diff
@@ -9493,12 +9501,20 @@ build_repo() {
...
+	    umount -f ${MASTERMNT}/packages
+	mount_packages
+
 	mkdir -p ${MASTERMNT}/tmp/packages
 	if [ -n "${PKG_REPO_SIGNING_KEY}" ]; then
 		msg "Signing repository with key: ${PKG_REPO_SIGNING_KEY}"
 		install -m 0400 "${PKG_REPO_SIGNING_KEY}" \
 			"${MASTERMNT:?}/tmp/repo.key"
 		injail ${PKG_BIN:?} repo \
+			${PKG_REPO_FLAGS} \
```
bdrewery, line 9517: A goal I have is for `set -u` to work. Please use `${PKG_REPO_FLAGS-}` here and in the next places.

src/share/poudriere/common.sh
bdrewery: Although I think you can just use `remount_packages`.

bdrewery (changes requested):
Quick changes requested

bdrewery:
@allanjude I can make the suggested changes myself if you are too busy. Want me to?

allanjude:
> @allanjude I can make the suggested changes myself if you are too busy. Want me to?

If you have time that would be great, I'd really like to get this landed so FreeBSD can switch to using it, so we can have per-release pkg repos without having to grow the mirrors

bdrewery:
Porting to 3.4.4 for next mini release. Will get it out much sooner.

PR #752, opened 2020-04-14
Use find(1)'s builtin -delete rather than forking rm

allanjude:
We might consider adding a -P ${PREPARE_PARALLEL_JOBS} to the xargs
to delete the files in parallel to speed things up.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -555,8 +555,7 @@ do_confirm_delete() {
 	if [ "${answer}" = "yes" ]; then
 		msg_n "Removing files..."
 		cat ${filelist} | tr '\n' '\000' | \
-		    xargs -0 -J % \
-		    find % -mindepth 0 -maxdepth 0 -exec rm -rf {} +
+		    xargs -0 rm -rf
 		echo " done"
 		ret=1
 	fi


src/share/poudriere/common.sh
```diff
@@ -555,8 +555,7 @@ do_confirm_delete() {
 	if [ "${answer}" = "yes" ]; then
 		msg_n "Removing files..."
 		cat ${filelist} | tr '\n' '\000' | \
-		    xargs -0 -J % \
-		    find % -mindepth 0 -maxdepth 0 -exec rm -rf {} +
+		    xargs -0 find % -mindepth 0 -maxdepth 0 -delete
```
bdrewery, line 558: I wonder why I have -mindepth -maxdepth 0. If this has a directory name in it will -delete recurse even with the maxdepth? I think not. In `logclean` I do only list directories that should be deleted by this function:
```
# Find build directories older than DAYS
BUILDNAME_GLOB="${BUILDNAME_GLOB}" SHOW_FINISHED=1 \
    for_each_build echo_logdir | \
    xargs -0 -J {} \
    find -x {} -type d -mindepth 0 -maxdepth 0 -Btime +${DAYS}d \
    > "${OLDLOGS}"
```

Also why remove the `-J %`?

allanjude: Removing the -J % was a mistake, it won't work without it.

The -mindepth 0 -maxdepth 0 means it will not traverse at all, it will only delete the files specifically listed in the % argument

bdrewery: I don't know why I didn't just `xargs -0 rm -rf`

allanjude: I had changed it to that, that is how the -J % got lost, but then changed it back because I figured there might have been a reason, and didn't want to make a subtle change to what it does

bdrewery:
```
# mkdir -p a/b/c/d/e
# find . -maxdepth 1 -mindepth 1 -name a
./a
# find . -maxdepth 1 -mindepth 1 -name a -delete
# find a
a
a/b
a/b/c
a/b/c/d
a/b/c/d/e
```

```
# echo a | xargs -J % find % -mindepth 0 -maxdepth 0
a
# echo a | xargs -J % find % -mindepth 0 -maxdepth 0 -delete
# find a
a
a/b
a/b/c
a/b/c/d
a/b/c/d/e
```

```
# echo a | xargs -J % find % -mindepth 0 -maxdepth 0 -exec rm -rf {} +
# find a
find: a: No such file or directory
```

bdrewery:
```
# mkdir -p a/b/c/d/e
# echo a | xargs rm -rf
# find a
find: a: No such file or directory
```
I think it can just be `xargs -0 rm -rf` instead of the xargs...find stuff.

allanjude: if you drop the -mindepth 0 -maxdepth 0, find - delete will work, but yeah, rm -rf is just as good

allanjude: What are your thoughts on using ${PREPARE_PARALLEL_JOBS} to run multiple rm's concurrently to delete the files more quickly?

bdrewery:
Thanks. I think this is good. Will test it tomorrow.

src/share/poudriere/common.sh
bdrewery: I missed your question before. If you think concurrent helps then sure.

bdrewery:
@allanjude This is good. Can you please fix the commit message though as we moved away from the `find -delete` change.

allanjude:
Updated

bdrewery (approved):

bdrewery:
Woops had the wrong merge message. That matters less though. Thanks!

PR #753, opened 2020-05-12
Clean up the svn/git repository after updating a jail to a newer vers…

rene0:
…ion.

Updating a jail to a newer version grows the .svn / .git metadata folder.
Compact it after updating and before taking a new @clean snapshot to
keep the size down.

It is unlikely that people will modify these checkouts by themselves. A
patch can be used for this when using svn and git supoprts feature branches
for this.

(WIP) Perhaps this needs a new knob in poudriere.conf.
(WIP) Perhaps we want to be more aggressive in the cleanup method.

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -580,6 +580,7 @@ install_from_vcs() {
 			msg_n "Updating the sources with ${METHOD}..."
 			${SVN_CMD} upgrade ${SRC_BASE} 2>/dev/null || :
 			${SVN_CMD} -q update -r ${TORELEASE:-head} ${SRC_BASE} || err 1 " fail"
+			${SVN_CMD} cleanup ${SRC_BASE} || err 1 " fail"
 			echo " done"
 			;;
 		git*)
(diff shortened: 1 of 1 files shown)


src/share/poudriere/jail.sh
```diff
@@ -580,13 +580,15 @@ install_from_vcs() {
...
 			${SVN_CMD} -q update -r ${TORELEASE:-head} ${SRC_BASE} || err 1 " fail"
+			${SVN_CMD} cleanup ${SRC_BASE} || err 1 " fail"
 			echo " done"
 			;;
 		git*)
 			${GIT_CMD} -C ${SRC_BASE} pull --rebase -q || err 1 " fail"
 			if [ -n "${TORELEASE}" ]; then
 				${GIT_CMD} checkout -q "${TORELEASE}" || err 1 " fail"
 			fi
+			(cd ${SRC_BASE} && ${GIT_CMD} gc) || err 1 " fail"
```
mat813, line 591: I'm not sure this is necessary, git will automatically run gc when it needs to.

rene0: Hmm, yes, it might be unnecessary.

bdrewery: I keep wanting to avoid the kitchen sink of options in Poudriere. It's already bad enough. This could work in your update script right?
```
    poudriere jail -u
    zfs destroy snap@clean
    (cd & &git gc)
    zfs snapshot snap@clean
```
Someone else may want to just drop all commits before HEAD too to save space. Or they just don't care and want the update to be as fast as possible.

rene0: Ah yes, this mini-script could work too. Such a cleanup should not take much time compared to buildworld, but the drop-everything-before-HEAD case is yet another option that one might or might not want.

PR #754, opened 2020-05-20
Fix permissions on the .json files

brd:
Without this the webUI does not work.

I don't really like this, but `mktemp` does not seem to have a way to set the perms when it creates the file.  I suppose we could use `umask` instead.. Thoughts?

### DIFF
--- a/src/share/poudriere/include/util.sh
+++ b/src/share/poudriere/include/util.sh
@@ -993,6 +993,7 @@ write_cmp() {
 	fi
 
 	if ! cmp -s "${dest}" "${tmp}"; then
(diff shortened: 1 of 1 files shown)


bdrewery:
I think if we change the `mktemp -t` to `mktemp -ut` it will work. It won't fix the newly broken files but at least umask will be respected again.

bdrewery:
We can use `chmod =rw $file` instead which will respect umask and avoid the `unlink`. https://unix.stackexchange.com/a/651303/7345. Helps that `chmod` can be made into a builtin too.

PR #756, opened 2020-06-10
image: Add pkgbase support

evadot:
When building a jail with -m src or using svn/git/etc ... if the user
adds -B it will make packages after buildworld/installworld.
The packages will be placed in ${POUDRIERE_DATA}/image/${JAILNAME}-repo.
When building a poudriere image the user can specify which pkgbase packages
will be installed in the resulting image with -P

Sponsored-by: The FreeBSD Foundation
Signed-off-by: Emmanuel Vadot <manu@FreeBSD.org>

### DIFF
--- a/src/share/poudriere/image.sh
+++ b/src/share/poudriere/image.sh
@@ -42,6 +42,7 @@ Parameters:
     -n imagename    -- The name of the generated image
     -o outputdir    -- Image destination directory
     -p portstree    -- Ports tree
+    -P pkgbase     -- List of pkgbase packages to install
     -s size         -- Set the image size
     -S snapshotname -- Snapshot name
     -t type         -- Type of image can be one of (default iso+zmfs):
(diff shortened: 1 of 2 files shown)


src/share/poudriere/image.sh
```diff
@@ -476,14 +485,35 @@ zsnapshot)
...
+	msg "mnt: ${mnt}"
+	OSVERSION=$(awk -F '"' '/REVISION=/ { print $2 }' ${mnt}/usr/src/sys/conf/newvers.sh | cut -d '.' -f 1)
+	msg "osversion: ${OSVERSION}"
+	mkdir -p ${WRKDIR}/world/etc/pkg/
+	cat << EOF > ${WRKDIR}/world/etc/pkg/FreeBSD-base.conf
+local: {
+        url: file://${POUDRIERE_DATA}/images/${JAILNAME}-repo/FreeBSD:${OSVERSION}:${arch}/latest,
+        enabled: true
+}
+EOF
```
bdrewery, line 500: You can do `-EOF` to still use tabs on all of these lines and the tabs will be eaten.

PR #761, opened 2020-07-03
image.sh: Support pre and post build scripts

allanjude:
This allows the user to override the default partitioning and other options without modifying poudriere

the -B <script> replaces the normal:
truncate
mdconfig
gpart

with the sourced script.

the -A <script> runs before the $WRKDIR/world is unmounted, allowing the contents of the image to be customized


src/share/poudriere/image.sh
```diff
@@ -734,6 +756,11 @@ zsnapshot)
 	;;
 esac
 
+if [ -f "${POST_BUILD_SCRIPT}" ]; then
+	# Source the pre-export-script.
+	. ${POST_BUILD_SCRIPT}
```
bdrewery, line 761: Missed some quotes here but the rest was fine.

bdrewery (approved):

PR #764, opened 2020-07-14
make svn+https default for ports

swills:


bdrewery:
For context this is for removing `portsnap` from official FreeBSD docs and deprecating it. Will need to clone this over to release-3.3 too. I can do that later.

PR #766, opened 2020-07-27
Add support for fetching from artifact.ci.freebsd.org

evadot:
This adds a new method -m freebsdci which fetches the latest tested sets
from the CI artifacts.
Sometimes head or stable doesn't compile and we waste time and power starting
building a jail to see later that it fails.
This also make it possible to create an armv7 jail from latest tested head which
isn't possible using the ftp method as re@ doesn't publish sets.

Signed-off-by: Emmanuel Vadot <manu@FreeBSD.org>

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -58,8 +58,8 @@ Options:
     -m method     -- When used with -c, overrides the default method for
                      obtaining and building the jail. See poudriere(8) for more
                      details. Can be one of:
-                       allbsd, ftp-archive, ftp, git, http, null, src=PATH, svn,
-                       svn+file, svn+http, svn+https, svn+ssh, tar=PATH
+                       allbsd, ftp-archive, ftp, freebsdci, git, http, null, src=PATH,
+                       svn, svn+file, svn+http, svn+https, svn+ssh, tar=PATH
                        url=SOMEURL.
     -P patch      -- Specify a patch to apply to the source before building.
     -S srcpath    -- Specify a path to the source tree to be used.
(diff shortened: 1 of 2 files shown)


src/share/poudriere/jail.sh
```diff
@@ -740,6 +741,7 @@ install_from_ftp() {
 				;;
 			allbsd) URL="https://pub.allbsd.org/FreeBSD-snapshots/${ARCH%%.*}-${ARCH##*.}/${V}-JPSNAP/ftp" ;;
 			ftp-archive) URL="http://ftp-archive.freebsd.org/pub/FreeBSD-Archive/old-releases/${ARCH%%.*}/${ARCH##*.}/${V}" ;;
+			freebsdci) URL="https://artifact.ci.freebsd.org/snapshot/${V}/latest_tested/${FULLARCH%%.*}/${FULLARCH##*.}" ;;
```
bdrewery, line 744: Everywhere I see `FULLARCH` it appears that `ARCH` suffices fine. The line right above this has the same expansions as this line.

evadot: No, ARCH is set to only the first componant if we give TARGET.TARGET_ARCH which we need to do as both are in the url in the artifact urls.
I should add that to the man page.

evadot: I think that the allbsd case doesn't work anymore.

evadot: Mhm, wait, I might have misread the code.

evadot: Yep sorry, fixed

evadot:
@bdrewery ok for you now ?

Issue #767, opened 2020-08-09
Off-by-one error of color change in output of `bulk` and `testport` command


yasuhirokimura:
OS: FreeBSD 12.1-RELEASE amd64
Poudriere: 3.2.4

There is off-by-one error of color change in output of `bulk` and `testport` command. It happens in line that reports build of one package is completed. Attached file is screenshot of poudriere output
on system cosole.

![poudriere off-by-one-error](https://user-images.githubusercontent.com/5249848/89742868-a4bf1a00-dad8-11ea-9e28-fd17a40e38b2.png)

Let's look at the line that reports build of `ports-mgmt/pkg` has completed.  The color of this line changes as following.

```
123456789012345678901234567890123456789012345678901234567890123456789012
[00:01:05] [01] [00:00:50] Finished ports-mgmt/pkg | pkg-1.14.6: Success
GGGGGGGGGGKGRRWKWWWWWWWWWWKGGGGGGGGKBBBBBBBBBBBBBBKBKBBBBBBBBBBBKGGGGGGG
```

Where,

W: white
R: red
G: green
B: blue
K: black

The error happens at 12th character from left ('['). Its color is green but probably should be white. Same error happens on every line that ports build of one package is completed.

yasuhirokimura:
I investigated the code and found there is bug in line 121 of `src/share/poudriere/common.sh`. That is,

```shell
                printf "${COLOR_ARROW}${elapsed}${DRY_MODE-}${arrow:+${COLOR_ARROW}${arrow}${COLOR_RESET} }${*}${COLOR_RESET}${NL}"
```

But how it should be fixed depends on what color is used to write the value of `${elapsed}`. If default color should be used, then this line should be changed as following.

```shell
                printf "${elapsed}${DRY_MODE-}${arrow:+${COLOR_ARROW}${arrow}${COLOR_RESET} }${*}${COLOR_RESET}${NL}"
```

And if `${COLOR_ARROW}` should be used. then it should be changed as following.

```shell
                printf "${COLOR_ARROW}${elapsed}${COLOR_RESET}${DRY_MODE-}${arrow:+${COLOR_ARROW}${arrow}${COLOR_RESET} }${*}${COLOR_RESET}${NL}"
```

yasuhirokimura:
Considering the result of comparing lines 121 and 123, it is inferred that the value of the variable `elapsed` should be displayed in the default color. So line 123 should be fixed as the former of previous comment.

[...]

PR #768, opened 2020-08-10
Fix color change error of message displayed when build of one package is finished.

yasuhirokimura:
Fix color change error of message displayed when build of one package is finished.

This fixes issue #767.


yasuhirokimura:
@bdrewery Updated.

PR #769, opened 2020-08-13
Adding new bulk command option: -W to mount ports tree in read-write …

ocochard:
…mode,

allowing to patch the port tree in interactive mode.

Interactive mode is used to troubleshoot a port build, so being able to use a `make makepatch` from it seems useful.

### DIFF
--- a/src/bin/poudriere-bulk.8
+++ b/src/bin/poudriere-bulk.8
@@ -243,6 +243,8 @@ The
 .Va WRKDIR
 will be tarred up into
 .No ${ Ns Va POUDRIERE_DATA Ns } Ns Pa /wrkdirs .
+.It Fl W
+Mount ports tree in read-write mode, allowing to patch it in interactive mode.
 .It Fl z Ar set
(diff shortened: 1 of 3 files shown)


bdrewery:
Thanks.

I'm conflicted by things like this as I want to steer Poudriere to be simpler which means less options. The `-i` flag feels (no data) obscure already and this added on is even more obscure.

Maybe it makes sense to always have the ports tree read-write when in interactive mode?
Or to have Poudriere install a script that you can run to do it - meaning a big notice when dropping to shell about interacting with the environment with Poudriere. The script would have to interact with some socket API to do it. I don't think we can delegate permissions to change the mount to the jail since it is nullfs but maybe.

ocochard:
> Maybe it makes sense to always have the ports tree read-write when in interactive mode?

Yes, if this option could be done it will be great (and simpler!)

ocochard:
I will propose a new Pull Request with "always have the ports tree read-write when in interactive mode".

PR #770, opened 2020-08-14
In interactive mode mount the ports tree in read-write mode, allowing…

ocochard:
… to patch it.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -1876,7 +1876,8 @@ do_portbuild_mounts() {
 	_pget portsdir ${ptname} mnt || err 1 "Missing mnt metadata for portstree"
 	[ -d ${portsdir}/ports ] && portsdir=${portsdir}/ports
 	${msgmount} "Mounting ports from: ${portsdir}"
-	${NULLMOUNT} -o ro ${portsdir} ${mnt}${PORTSDIR} ||
+	[ ${INTERACTIVE_MODE} -gt 0 ] && OPT=rw || OPT=ro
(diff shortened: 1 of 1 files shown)


bdrewery (commented):
Thanks for the fix and sorry for the delay.

It would be safer to only do this in `enter_interactive` with remounting.
It's more complicated than I anticipated because nullfs does not support `mount -u -w` apparently.
All of this code (from `do_portbuild_mounts`) may need to be reran

```
_pget portsdir ${ptname} mnt || err 1 "Missing mnt metadata for portstree"    
[ -d ${portsdir}/ports ] && portsdir=${portsdir}/ports                        
${msgmount} "Mounting ports from: ${portsdir}"                                
${NULLMOUNT} -o ro ${portsdir} ${mnt}${PORTSDIR} ||                           
        err 1 "Failed to mount the ports directory "                          
for o in ${OVERLAYS}; do                                                      
        _pget odir "${o}" mnt || err 1 "Missing mnt metadata for overlay ${o}"
        ${msgmount} "Mounting ports overlay from: ${odir}"                    
        ${NULLMOUNT} -o ro "${odir}" "${mnt}${OVERLAYSDIR}/${o}"              
done                                                                          
${msgmount} "Mounting packages from: ${PACKAGES_ROOT}"                        
mount_packages -o ro                                                          
${msgmount} "Mounting distfiles from: ${DISTFILES_CACHE}"                     
${NULLMOUNT} ${DISTFILES_CACHE} ${mnt}/distfiles ||                           
        err 1 "Failed to mount the distfiles cache directory"                 
```

We can unmount all of that with
`umountfs ${MASTERMNT}/${PORTSDIR} 1`

That's getting too complicated and error prone. I'll just merge this for now.
We can unmount the ports dir and remount it.

Issue #78, opened 2014-09-18
better error exceptions


bapt:
I have seen a some strange errors like "====>> Dependency loop or poudriere bug detected." right now there is even with the DEBUG mode no chance to detect what exactly the error is. I would like to request to rethink about the error exceptions handling. It must be possible to write a sort of debug log where you are able to see whats going on.

<hr /><i>bdrewery added on 2013-03-11 10:41:15 UTC:</i><br />
This is addressed now. The unprocessable queue errors are smarter and point more to the problems.

PR #772, opened 2020-08-18
Allow overlay to reinstate a port.

web-vertalo:
This update allows an overlay to reinstate a port with a MOVED entry. It applies the following convention to reinstate a deleted origin:

`old name (blank for reinstated)|new name|date of move|reason`

MOVED files from overlays are effectively appended to the MOVED file from the base repo. This is possibly incorrect for any use except simple reinstatement.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -7060,8 +7060,13 @@ load_moved() {
 	[ -f ${MASTERMNT}${PORTSDIR}/MOVED ] || return 0
 	msg "Loading MOVED for ${MASTERMNT}${PORTSDIR}"
 	bset status "loading_moved:"
-	awk -f ${AWKPREFIX}/parse_MOVED.awk \
-	    ${MASTERMNT}${PORTSDIR}/MOVED | \
+        { cat ${MASTERMNT}${PORTSDIR}/MOVED 
+          for o in ${OVERLAYS}; do
+                 test -f "${MASTERMNT}${OVERLAYSDIR}/${o}/MOVED" || continue
+                 cat "${MASTERMNT}${OVERLAYSDIR}/${o}/MOVED"
+          done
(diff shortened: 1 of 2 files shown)


src/share/poudriere/common.sh
```diff
@@ -7060,8 +7060,13 @@ load_moved() {
 	[ -f ${MASTERMNT}${PORTSDIR}/MOVED ] || return 0
 	msg "Loading MOVED for ${MASTERMNT}${PORTSDIR}"
 	bset status "loading_moved:"
-	awk -f ${AWKPREFIX}/parse_MOVED.awk \
-	    ${MASTERMNT}${PORTSDIR}/MOVED | \
+        { cat ${MASTERMNT}${PORTSDIR}/MOVED
```
bdrewery, line 7063: Needs a `local o` at the top of the function to avoid leaking `$o`

bdrewery, line 7063: We can avoid calling `cat` with a variable. And should fixup quoting since touching this.
```
local movedfiles

movedfiles="${MASTERMNT}${PORTSDIR}/MOVED"
for o in ${OVERLAYS}; do
                 test -f "${MASTERMNT}${OVERLAYSDIR}/${o}/MOVED" || continue
                 movedfiles="${movedfiles} ${MASTERMNT}${OVERLAYSDIR}/${o}/MOVED"
done
awk -f ${AWKPREFIX}/parse_MOVED.awk ${movedfiles} | \
...
```

I doubt we would hit the command line argument limit here but if we do then this pattern would be more efficient
```
{
echo file1
for o; do
echo $o
done
} | xargs cat | awk -f ...
```

Best to avoid fork/exec if possible.

web-vertalo:
Replaced by https://github.com/freebsd/poudriere/pull/78

PR #774, opened 2020-08-21
A wrapper script for developer debugging of processonelog.sh.

linimon:
Suggested use: download a directory's worth of errorlogs and run this script with the default (checked-in) processonelog.sh.  Then copy the resulting regresslogs.out to regresslogs.out.wanted and edit the results to taste.  (file format: "<errorlogname>: <reason>"). Then edit a local copy of processonelog.sh, switch this script to use it, and run it again.  You will be able to check your results in diff.out.


linimon:
processonelog.sh: update many more patterns to attempt to minimize "???" as much as possible.  While here, remove no-longer-used Reasons in interest of speed.  (In fact, the newest version is faster than before.)
regresslogs.sh: add new script used to regression-test changes to processonelog.sh.

linimon:
Will issue new pull request to update this one.

PR #777, opened 2020-08-30
processonelog.sh: update many patterns, add test harness

linimon:
processonelog.sh: cumulative update of multiple patterns, to attempt to eliminate as many occurrences of "???" as feasible.
While here, remove no-longer-seen Reasons for performance.  (In fact, the latest version of the script is actually faster.)

regresslogs.sh: add a test harness for working with processonelog.sh.  This is a developer-only script and is actually not used in mainline poudriere.


linimon:
Superceded by 782.

PR #781, opened 2020-09-05
Processonelog corrected

linimon:
Catch up with recent cleanups and additions to processonelog.sh.


linimon:
I will keep guessing what the UI wants.

PR #782, opened 2020-09-05
Catch up with all recent changes to processonelog.sh.

linimon:


bdrewery:
@linimon This is ready to merge as is right?

linimon:
> @linimon This is ready to merge as is right?

This has been ready since 20200905.

PR #785, opened 2020-09-10
Always use fully qualified arch in test suite jail creation.

bdragon28:
This allows me to successfully run `make check` on powerpc.powerpc64. (Without this, it tries to download the base system from an invalid path.)


bdrewery (approved):
Thanks for running the test suite :)

PR #786, opened 2020-09-14
Add new pkg repo file hash ability to poudriere.conf Default to off

darkfiberiru:
@allanjude  Here's an initial addition to poudriere. Do you think it makes sense to expose both hash and symlink ability or just default to creating symlink if your appending hash to package name

If anyone has a prefered method to a poudriere.conf variable Please let me know

@allanjude  Also any help on wording for poudriere.conf comment would be appreciated

### DIFF
--- a/src/etc/poudriere.conf.sample
+++ b/src/etc/poudriere.conf.sample
@@ -309,3 +309,13 @@ DISTFILES_CACHE=/usr/ports/distfiles
 # processing of the queue slightly, especially for bulk -a builds.
 # Default: no
 #HTML_TRACK_REMAINING=yes
+
+# Have pkg create hashed versions of the pkg filenames with symlinks to
+# original pkg names. The packagesite.yaml file will point to the hashed version
+# of these files. By using hashed pkg filenames, this allows users to lazily
+# synchronise packages without conflicting with the current packages,
+# for example using rsync or CDNs.  Once the packages are synced the much
+# smaller meta files can then be synced. Allowing a near atomic update of repo.
+# On caching cdn this means a need to purge 2-5 files instead of all pkgs that
+# have been updated.
+#PKG_HASH=no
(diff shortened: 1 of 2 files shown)


swills:
Is it me or is there a mix-up between `$PKG_OPTIONS` and `$REPO_OPTIONS` here? Looks like `$PKG_OPTIONS` is set and not used and `$REPO_OPTIONS` is used but not set.

src/etc/poudriere.conf.sample
```diff
@@ -309,3 +309,9 @@ DISTFILES_CACHE=/usr/ports/distfiles
 # processing of the queue slightly, especially for bulk -a builds.
 # Default: no
 #HTML_TRACK_REMAINING=yes
+
+# Have pkg create hashed versions of the pkg filenames with symlinks to
+# original pkg names. The packagesite.yaml file will point to the hashed version
+# of these files. This is usefaully in a cdn enviroment to ensure pkg names are
```
dch, line 315: s/faully/ful/ && s/enviroment/environment/

Maybe rephrase this comment as:

By using hashed pkg filenames, this allows users to lazily synchronise packages without conflicting with the current packages, for example using rsync or CDNs.


I must admit I'm not clear on the exact steps at the end to switch the package repos over, but maybe this doesn't matter in practice?

darkfiberiru: Thanks. Sorry i shouldn't code at 2am but I just wanted to get a review started for feedback if I was even doing this the right way.

darkfiberiru (commented):
@swills Thank you indecisiveness striked on variable naming and I didn't proofread well enough. Fixed now along with improvements to pourdriere.conf option comment.  @dch Thanks for input. Does the new wording read any better.

darkfiberiru:
#751  Apparently this is a duplicate as I didn't search prior june when the pkg hash work was redone and allan had submitted a PR in april

Issue #740, opened 2020-01-29
Testport fails to run on an origin from an overlay


0mp:
So I have the following scenario:

1. `/usr/ports` is where my ports tree is located.
2. `/opt/overlay` is where my overlay is located. It has only one port (`/opt/overlay/devel/myport`).

Here's what I do:

```script
# poudriere ports -c -p default -m null -M /usr/ports
# poudriere ports -c -p overlay -m null -M /opt/overlay
```

Now, `poudriere bulk -j myjail -O overlay devel/myport` works just fine, whereas `poudriere testport -j myjail -O overlay -o devel/myport` does not, failing with:

```
[00:00:03] Error: Nonexistent origin devel/myport
```

I took a look into `testport.sh` and I found the following lines: https://github.com/freebsd/poudriere/blob/master/src/share/poudriere/testport.sh#L208-L210

I am not sure how hard it would be to support running testport on origins from an overlay.

bdrewery:
Yeah the whole file needs to learn about overlays apparently.

For now just use `bulk -t` instead.

mandree:
related: #352

bdrewery:
Fixed by #790

PR #790, opened 2020-10-20
Improve overlay support

brooksdavis:
Allow testport to work with an origin in an overlay.

Tighten the parameters on selecting an overlay directory as an origin.


bdrewery:
Fixes #740

0mp:
That's great news! I'll give it a shot soon!

PR #792, opened 2020-10-22
Add an example of MAKEWORLDARGS

brooksdavis:
It took me quite a bit of grepping to find this otherwise undocumented feature. We might even consider enabling the example settings by default.


grahamperrin:
If `WITH_MALLOC_PRODUCTION=yes`, then what is the effect of _also_ having `-DMALLOC_PRODUCTION`?

(I see things such as <https://github.com/freebsd/freebsd-src/commit/86b019c29a21c2664ccb5d59ee0c999edecd23b6>, but don't understand.)

PR #793, opened 2020-10-23
jail: rework command selection

brooksdavis:
Set a command name rather than creating a pseudo-bitmask that requires
each case to be updated each time a command is added.

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -1030,21 +1030,18 @@ info_jail() {
 
 get_host_arch ARCH
 REALARCH=${ARCH}
-START=0
-STOP=0
-LIST=0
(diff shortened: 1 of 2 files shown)


bdrewery:
I think this is fine but would prefer all uses of it are changed at once. Other subcommands have similar patterns.

brooksdavis:
The only other one I spotted was ports (and it at least didn't use the pseudo bit mask). I've updated it.

I'd personally like to change these commands to work more like `poudriere jail create <options>` and with this change as a basis that could easily be done in a backwards compatible way.

bdrewery:
@brooksdavis Thanks for doing this. Yes I am open to a new CLI but I think it needs to be planned in advance better.

PR #795, opened 2020-10-26
jail: Add archive subcommand and create method

brooksdavis:
The archive subcommand (-A) creates a tar file containing the contents of a jail and it's metadata.  The create subcommand (-c) gains a new method to create a new jail from an archive, restoring its metadata. This can be used to save a copy of a jail before updates or to transfer a jail to another machine.

This PR depends on (and thus contains) #793


bdrewery:
@brooksdavis Can you please rebase after the #793 merge? Looks like the ports piece was missed here.

PR #797, opened 2020-11-03
bulk: allow prefetching binary packages from official repos

bapt:
As soon as we have the entire list of packages to build but before
the sanity check, pre fetch all the binary packages that we can
from the official repositories.

The sanity check running after than will probably destroy some
packages for which the confiration will be different or any reason
suitable for the sanity check to destroy packages.

When listing the packages to fetch only list the one not already
existing to avoid pkg fetch to redownload and overwrite packages
we have already customized

### DIFF
--- a/src/bin/poudriere-bulk.8
+++ b/src/bin/poudriere-bulk.8
@@ -40,6 +40,7 @@
 .Fl a
 .Fl j Ar name
 .Op Fl CcFIikNnRrSTtvw
+.Op Fl b Ar name
 .Op Fl B Ar name
 .Op Fl J Ar maxjobs Ns Op Cm \&: Ns Ar prebuildmaxjobs
 .Op Fl O Ar overlays
@@ -50,6 +51,7 @@
 .Fl f Ar file Op Oo Fl f Ar file2 Oc Ar ...
 .Fl j Ar name
 .Op Fl CcFIikNnRrSTtvw
+.Op Fl b Ar name
 .Op Fl B Ar name
 .Op Fl J Ar maxjobs Ns Op Cm \&: Ns Ar prebuildmaxjobs
 .Op Fl O Ar overlays
@@ -59,6 +61,7 @@
 .Cm bulk
 .Fl j Ar name
 .Op Fl CcFIikNnRrSTtvw
+.Op Fl b Ar name
 .Op Fl B Ar name
 .Op Fl J Ar maxjobs Ns Op Cm \&: Ns Ar prebuildmaxjobs
 .Op Fl O Ar overlays
@@ -111,6 +114,17 @@ arguments may be specified at once.
 .El
 .Sh OPTIONS
 .Bl -tag -width "-B name"
+.It Fl b Ar name
+Specify the
+.Ar name
+of the binary package branch to use to prefetch packages.
+Should be
+.Qq latest
+or
+.Qq quarterly .
+With this option poudriere will first try to fetch from the binary package
+repository specified the binary package prior to do the sanity check if the
+package does not already exist.
 .It Fl B Ar name
 Specify which buildname to use.
 By default
--- a/src/bin/poudriere-bulk.8
+++ b/src/bin/poudriere-bulk.8
@@ -141,6 +141,13 @@ is used along with
 or
 .Fl c ,
 then listed packages will not be fetched.
+.Pp
+See
+.Sy PACKAGE_FETCH_BRANCH ,
+and
+.Sy PACKAGE_FETCH_URL ,
+in
+.Pa poudriere.conf.sample .
 .It Fl C
 Clean only the packages specified on the command line or in the file given by
 .Fl f Ar file .
--- a/src/bin/poudriere-bulk.8
+++ b/src/bin/poudriere-bulk.8
@@ -119,9 +119,10 @@ Specify the
 .Ar name
 of the binary package branch to use to prefetch packages.
 Should be
-.Qq latest
+.Qq latest ,
+.Qq quarterly ,
 or
-.Qq quarterly .
+.Qq release_* .
 With this option poudriere will first try to fetch from the binary package
 repository specified the binary package prior to do the sanity check if the
 package does not already exist.
--- a/src/bin/poudriere-bulk.8
+++ b/src/bin/poudriere-bulk.8
@@ -114,6 +114,13 @@ arguments may be specified at once.
 .El
 .Sh OPTIONS
 .Bl -tag -width "-B name"
+.It Fl B Ar name
+Specify which buildname to use.
+By default
+.Ar YYYY-MM-DD_HH:MM:SS
+will be used.
+This can be used to resume a previous build and use the same log and URL paths.
+Resuming a build will not retry built/failed/skipped/ignored packages.
 .It Fl b Ar name
 Specify the
 .Ar name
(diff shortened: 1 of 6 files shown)


src/bin/poudriere-bulk.8
```diff
@@ -111,6 +114,14 @@ arguments may be specified at once.
 .El
 .Sh OPTIONS
 .Bl -tag -width "-B name"
+.It Fl b Ar name
+Specify the name of the binary package to use to prefetch packages. Should be
```
brooksdavis, line 118:
```suggestion
Specify the name of the binary package branch to use to prefetch packages.
Should be
```

0mp, line 118: "binary package" is a typo, right?

brooksdavis (commented):
This looks super helpful for incremental builds on cloud-based poudriere servers.

src/share/poudriere/bulk.sh
```diff
@@ -92,14 +94,22 @@ OVERLAYS=""
...
 		a)
 			ALL=1
 			;;
 		B)
 			BUILDNAME="${OPTARG}"
 			;;
+		b)
+			PACKAGE_BRANCH="${OPTARG}"
+			case "${PACKAGE_BRANCH}" in
+			latest|quaterly) ;;
```
kevans91, line 108:
```suggestion
			latest|quarterly) ;;
```

src/share/poudriere/bulk.sh
```diff
@@ -92,14 +94,22 @@ OVERLAYS=""
...
+			*)
+				err 1 "Invalid branch rename for packages: ${OPTARG}"
```
kevans91, line 110:
```suggestion
				err 1 "Invalid branch name for packages: ${OPTARG}"
```

brooksdavis:
It would be nice if there was a way to specify a full URL rather than hardcoding pkg.freebsd.org. Supporting `latest|quarterly|<URL>` seems viable.

src/bin/poudriere-bulk.8
```diff
@@ -111,6 +114,17 @@ arguments may be specified at once.
...
+.It Fl b Ar name
+Specify the
+.Ar name
+of the binary package branch to use to prefetch packages.
+Should be
+.Qq latest
+or
+.Qq quarterly .
+With this option poudriere will first try to fetch from the binary package
+repository specified the binary package prior to do the sanity check if the
```
bdrewery, line 126: With this option poudriere will try to fetch binary packages from the specified repository prior to doing its sanity check on local packages.

src/bin/poudriere-bulk.8
```diff
@@ -111,6 +114,17 @@ arguments may be specified at once.
 .El
 .Sh OPTIONS
 .Bl -tag -width "-B name"
+.It Fl b Ar name
+Specify the
+.Ar name
+of the binary package branch to use to prefetch packages.
```
bdrewery, line 120: `-b` doesn't make me think fetch at all. In bulk.sh it is `-b branch` which makes me think 'branch' rather than 'binary'; I'm saying that at a quick glance of the manpage I might think this is saying which branch to build from rather than anything to do with fetching. I don't really have a better idea.
We're getting to the point of needing to use things like `--fetch`. I had been thinking `--seed` but `-f -F -S` are taken already and I had a plan to flip `-S` into `-s` IIRC.

Like Brooks comment I think this inevitably ends in someone asking for `REPO_URL` to be configurable so this flag interface may need to forward think about that.

src/share/poudriere/common.sh
```diff
@@ -2940,6 +2940,29 @@ jail_cleanup() {
 	export CLEANED_UP=1
 }
 
+download_from_repo() {
+	msg "Prefetching missing packages from pkg+http://pkg.freebsd.org/\${ABI}/${PACKAGE_BRANCH}"
```
bdrewery, line 2944:
```
local pkgname originspec _ignored
```

src/share/poudriere/common.sh
```diff
@@ -2940,6 +2940,29 @@ jail_cleanup() {
...
+        url: pkg+http://pkg.freebsd.org/\${ABI}/${PACKAGE_BRANCH};
+}
+EOF
+	umount ${UMOUNT_NONBUSY} ${MASTERMNT}/packages || \
+	    umount -f ${MASTERMNT}/packages
+	mount_packages
+	# only list packages which do not exists to prevent pkg from overwriting prebuilt packages
+	# XXX only work when PKG_EXT is the same as the upstream
+	(
+	while mapfile_read_loop "all_pkgs" pkgname originspec _ignored; do
```
bdrewery, line 2956: This assumes ` pwd`, please add this assertion at the top. Even if functionally it is guaranteed now I think the assertion is a good documentation of the subtle assumption.
```
[ "${PWD}" = "${MASTERMNT}/.p" ] || \                     
    err 1 "download_from_repo requires PWD=${MASTERMNT}/.p"
```

src/share/poudriere/common.sh
```diff
@@ -7371,6 +7394,10 @@ prepare_ports() {
 		fi
 	fi
 
+	if [ -n "${PACKAGE_BRANCH}" ]; then
+		download_from_repo
+	fi
```
bdrewery, line 7399: 1. Need this to not run for non-build commands (like `pkgclean`) and to not run when resuming, and to still respect `-c`  and `-C`. So I think it makes sense higher up (Github review won't let me comment on untouched lines.) :
```
if [ -n "${PACKAGE_BRANCH}" -a "${resuming_build}" -ne 1 ]; then
		download_from_repo
fi
# existing code
if [ ${JAIL_NEEDS_CLEAN} -eq 1 ]; then
			msg_n "Cleaning all packages due to newer version of the jail..."
			rm -rf ${PACKAGES}/* ${cache_dir}
```

2. Then there is the question of what should `-n` (dry run) do? Should it fetch so that a proper accounting of what will build is done or should it just say "will try fetching pkgA pkgB ..."? I have no opinion here.

src/share/poudriere/common.sh
```diff
@@ -2940,6 +2940,29 @@ jail_cleanup() {
...
+	umount ${UMOUNT_NONBUSY} ${MASTERMNT}/packages || \
+	    umount -f ${MASTERMNT}/packages
+	mount_packages
+	# only list packages which do not exists to prevent pkg from overwriting prebuilt packages
+	# XXX only work when PKG_EXT is the same as the upstream
+	(
+	while mapfile_read_loop "all_pkgs" pkgname originspec _ignored; do
+		[ -f ${MASTERMNT}/packages/All/${pkgname}.${PKG_EXT} ] || echo ${pkgname}
+	done
+	)| JNETNAME="n" injail xargs env -i ASSUME_ALWAYS_YES=yes pkg fetch -o /packages
```
bdrewery, line 2959: I think this is fine for now but ultimately we may want to inspect more what will be downloaded to avoid `delete_old_pkg` from nuking the same packages every time for things like different options or flavors.

bdrewery:
Remaining items I am fixing before merge:

- [ ]  pkg bootstrapping; pkg is fetched but then the build fails because symlink packages/Latest/pkg.txz is missing.
- [ ] bulk -n (dry mode) still is fetching
- [ ] bulk -C needs to only fetch dependencies and not the packages that are listed, or a second pass to delete listed ports
- [ ] bulk -c: It is unclear what this should do. Should it just mirror the remote or still build listed ports?

Items for after:
- [ ] #822 is likely to be a problem here too due to out-of-sync remote and local tree.
- [ ] Need to compare options *before* fetching or every build will likely fetch and delete which wastes time and bandwidth. 1 differing option could cause thousands of unneeded packages to be fetched.

PR #799, opened 2020-11-06
jail: pkgbase: Add update-packages support

evadot:
Now that head support update-packages add support in jail update.
Since pkgbase is present in 11/12/13 but update-packages is only in head
and will not be mfc add a fallback to make packages if update-packages rule
doesn't exists.
While here do not allow building packages if not kernel was provided.

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -992,6 +992,7 @@ info_jail() {
 	local elapsed elapsed_days elapsed_hms elapsed_timestamp
 	local now start_time timestamp
 	local jversion jarch jmethod pmethod mnt fs kernel
+	local pkgbase
 
 	jail_exists ${JAILNAME} || err 1 "No such jail: ${JAILNAME}"
 
@@ -1017,6 +1018,7 @@ info_jail() {
 	_jget mnt ${JAILNAME} mnt || :
 	_jget fs ${JAILNAME} fs || fs=""
 	_jget kernel ${JAILNAME} kernel || kernel=
+	_jget pkgbase ${JAILNAME} pkgbase || pkgbase=0
 
 	echo "Jail name:         ${JAILNAME}"
 	echo "Jail version:      ${jversion}"
@@ -1032,6 +1034,11 @@ info_jail() {
 	fi
 	if [ -n "${timestamp}" ]; then
 		echo "Jail updated:      $(date -j -r ${timestamp} "+%Y-%m-%d %H:%M:%S")"
+	fi
+	if [ "${pkgbase}" -eq 0 ]; then
+	    echo "Jail pkgbase:      disabled"
+	else
+	    echo "Jail pkgbase:      enabled"
 	fi
 	if [ "${PTNAME_ARG:-0}" -eq 1 ] && porttree_exists ${PTNAME}; then
 		_pget pmethod ${PTNAME} method
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -208,7 +208,37 @@ hook_stop_jail() {
 	fi
 }
 
+update_pkgbase() {
+	local make_jobs
+	local destdir="${JAILMNT}"
+
+	if [ ${JAIL_OSVERSION} -gt 1100086 ]; then
+		make_jobs="${MAKE_JOBS}"
+	fi
+
+	msg "Starting make update-packages"
+	${MAKE_CMD} -C "${SRC_BASE}" ${make_jobs} update-packages \
+	    DESTDIR="${destdir}" REPODIR="${POUDRIERE_DATA}/images/${JAILNAME}-repo" ${MAKEWORLDARGS}
+	case $? in
+	    0)
+		run_hook jail pkgbase "${POUDRIERE_DATA}/images/${JAILNAME}-repo"
+		return
+		;;
+	    2)
+		${MAKE_CMD} -C "${SRC_BASE}" ${make_jobs} packages \
+		    DESTDIR="${destdir}" REPODIR="${POUDRIERE_DATA}/images/${JAILNAME}-repo" ${MAKEWORLDARGS} || \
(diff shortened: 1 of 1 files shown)


src/share/poudriere/jail.sh
```diff
@@ -1210,6 +1243,9 @@ else
 	GIT_FULLURL=${proto}://${GIT_BASEURL}
 fi
 
+if [ -z "${KERNEL}" ] && [ "${BUILD_PKGBASE}" -eq 1 ]; then
+    err 1 "pkgbase build need a kernel"
```
bdrewery, line 1247: Why?

src/share/poudriere/jail.sh
```diff
@@ -355,6 +379,10 @@ update_jail() {
 		err 1 "Unsupported method"
 		;;
 	esac
+	pkgbase=$(jget ${JAILNAME} pkgbase)
+	if [ ! -z "${pkgbase}" ] && ${pkgbase} -eq 1 ]; then
```
bdrewery, line 383: There's a syntax error here. Missing `[` after `&&`.
Also should quote the `${pkgbase}`
And `-n` rather than `! -z` is more the style used.
And need a `local pkgbase` at the top.

src/share/poudriere/jail.sh
```diff
@@ -924,6 +954,8 @@ create_jail() {
 	jset ${JAILNAME} method ${METHOD}
 	[ -n "${FCT}" ] && ${FCT} version_extra
 
+	jset ${JAILNAME} pkgbase ${BUILD_PKGBASE}
```
bdrewery, line 957: Might as well extend `jail -i` to show this value in `info_jail`. Possibly `jail -l` too in `list_jail`

src/share/poudriere/jail.sh
```diff
@@ -208,6 +208,30 @@ hook_stop_jail() {
...
+	if [ ${JAIL_OSVERSION} -gt 1100086 ]; then
+		make_jobs="${MAKE_JOBS}"
+	fi
+
+	msg "Starting make update-packages"
+	${MAKE_CMD} -C "${SRC_BASE}" ${make_jobs} update-packages \
+		    DESTDIR=${destdir} REPODIR=${POUDRIERE_DATA}/images/${JAILNAME}-repo ${MAKEWORLDARGS}
+	if [ $? -eq 0 ]; then
+	    run_hook jail pkgbase ${POUDRIERE_DATA}/images/${JAILNAME}-repo
+	   return
```
bdrewery, line 224: Appears there is a whitespace issue at the beginning of one of these lines.

src/share/poudriere/jail.sh
```diff
@@ -208,6 +208,30 @@ hook_stop_jail() {
...
+	local make_jobs
+	local destdir="${JAILMNT}"
+
+	if [ ${JAIL_OSVERSION} -gt 1100086 ]; then
+		make_jobs="${MAKE_JOBS}"
+	fi
+
+	msg "Starting make update-packages"
+	${MAKE_CMD} -C "${SRC_BASE}" ${make_jobs} update-packages \
+		    DESTDIR=${destdir} REPODIR=${POUDRIERE_DATA}/images/${JAILNAME}-repo ${MAKEWORLDARGS}
```
bdrewery, line 221: Should quote all of this (and later)
`DESTDIR="${destdir}" REPODIR="..."`
Don't quote `${MAKEWORLDARGS}` though of course.

src/share/poudriere/jail.sh
```diff
@@ -208,6 +208,30 @@ hook_stop_jail() {
...
+	if [ $? -eq 0 ]; then
+	    run_hook jail pkgbase ${POUDRIERE_DATA}/images/${JAILNAME}-repo
```
bdrewery, line 223: Really most raw paths should be quoted.

src/share/poudriere/jail.sh
```diff
@@ -208,6 +208,30 @@ hook_stop_jail() {
...
+	   return
+	elif [ $? -eq 2 ]; then
```
bdrewery, line 225: This `$?` will always be 0 or 1 because it's the result of the `if [ $? -eq 0 ]` line.
```
# cat ret.sh
( exit 2 )
if [ $? -eq 0 ]; then
        echo 0
elif [ $? -eq 2 ]; then
        echo 2
else
        echo "lost ?"
fi
```
```
# sh ret.sh
lost ?
```
Some ways to avoid it:
```
ret=$?
if [ $ret ...
```
```
case $? in
0) # Success
 ...
 ;;
2) # Missing target, try fallback
  ...
  ;;
*) # Some other failure
  err 1 "Failed to 'make update-packages'"
  ;;
esac
```

src/share/poudriere/jail.sh
```diff
@@ -1045,6 +1077,7 @@ SETNAME=""
 XDEV=0
 BUILD=0
 GIT_DEPTH=--depth=1
+BUILD_PKGBASE=0
```
bdrewery, line 1080: How is this getting set? In Poudriere.conf? It should have an example there or should be an added cmdline option.

bdrewery (changes requested):

src/share/poudriere/jail.sh
evadot: Because make packages create world+kernel packages.
Also the main reason for using poudriere pkgbase is either to create poudriere images or to distribute packages in sync with kernel (like kmods).
If there is a demand for world only packages I'll add support later.

src/share/poudriere/jail.sh
evadot: I can do that in a later patch yes.

src/share/poudriere/jail.sh
evadot: with -B on the jail creation (already upstreamed).

src/share/poudriere/jail.sh
evadot: Done

PR #804, opened 2020-11-16
Fix default ports update method

### LINKED ISSUE #803: Documented and actual default methods (jail and ports) don't match
The [documented default jail method is "http"](https://github.com/freebsd/poudriere/blob/f4bf593e00161d2d901f5b24fd3bd38d7db7471b/src/bin/poudriere-jail.8#L197). The [actual default jail method is"ftp"](https://github.com/freebsd/poudriere/blob/f4bf593e00161d2d901f5b24fd3bd38d7db7471b/src/share/poudriere/jail.sh#L220).

The [documented default ports method is "svn+https"](https://github.com/freebsd/poudriere/blob/f4bf593e00161d2d901f5b24fd3bd38d7db7471b/src/bin/poudriere-ports.8#L111). The [actual default ports method is "portsnap"](https://github.com/freebsd/poudriere/blob/f4bf593e00161d2d901f5b24fd3bd38d7db7471b/src/share/poudriere/ports.sh#L362).

swills:
Fixes #803 partially

### DIFF
--- a/src/share/poudriere/ports.sh
+++ b/src/share/poudriere/ports.sh
@@ -27,6 +27,8 @@
 
 . ${SCRIPTPREFIX}/common.sh
 
+METHOD_DEF=svn+https
+
 # test if there is any args
 usage() {
 	cat << EOF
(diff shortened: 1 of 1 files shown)


bdrewery (changes requested):
This is good but let's move the default to 1 line so we don't run into it again.

Something like `METHOD_DEF=svn+https` up top and then replace this and line 134 with that var.
And the `usage` use of the default can be changed to the var too :)

swills:
> This is good but let's move the default to 1 line so we don't run into it again.
> 
> Something like `METHOD_DEF=svn+https` up top and then replace this and line 134 with that var.
> And the `usage` use of the default can be changed to the var too :)

Done, look OK?

bdrewery (approved):

PR #813, opened 2020-12-25
set the canonical FreeBSD base repo

igalic:
since the migration of base to git, we should use that as the canonical git repo

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -7851,7 +7851,7 @@ if [ -z "${NO_ZFS}" ]; then
 fi
 
 : ${SVN_HOST="svn.freebsd.org"}
-: ${GIT_BASEURL="github.com/freebsd/freebsd.git"}
+: ${GIT_BASEURL="git.freebsd.org/src.git"}
 : ${GIT_PORTSURL="github.com/freebsd/freebsd-ports.git"}
 : ${FREEBSD_HOST="https://download.FreeBSD.org"}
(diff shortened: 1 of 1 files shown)


igalic:
this doesn't actually seem to work…

```
meena@fbsd12-pkgbase-hel ~> sudo poudriere jail -c -j pkgbase130-current-amd64 -a amd64 -B -K GENERIC -b -m git -v main
[00:00:00] Creating pkgbase130-current-amd64 fs at /poudriere/jails/pkgbase130-current-amd64... done
[00:00:01] Checking out the sources with git...fatal: unable to connect to git.freebsd.org:
git.freebsd.org[0: 2604:1380:2000:9501::e6a:1]: errno=Connection refused
git.freebsd.org[1: 139.178.72.204]: errno=Connection refused

[00:00:01] Error:  fail
[00:00:01] Error while creating jail, cleaning up.
[00:00:01] Removing pkgbase130-current-amd64 jail... done
[00:00:02] Cleaning pkgbase130-current-amd64 data... done
```

igalic:
if this is failing, then perhaps i should change the default protocol for git as well, to https

bdrewery:
I ended up merging #817 which modified the github repo URL. The github repo supports more protocols, has a better CDN, and better maintenance/funding (security) than FreeBSD's repo. Beyond "it is canonical" I think it is more practical and user friendly to use the github one for now.

It could make sense to cleanup the sample config to provide options around this.

Issue #794, opened 2020-10-24
Unhardcode compression=lz4 when creating jails on ZFS


jbeich:
I'd like to compress jails more aggressively with gzip, zstd or newer but poudriere forces lz4.
```
$ zfs get compression tank/poudriere/jails
NAME                  PROPERTY     VALUE           SOURCE
tank/poudriere/jails  compression  zstd-18          local

$ poudriere jail -cj 114amd64 -a amd64 -v 11.4-RELEASE

$ zfs get compression tank/poudriere/jails/114amd64
NAME                           PROPERTY     VALUE           SOURCE
tank/poudriere/jails/114amd64  compression  lz4             local
```

igalic:
can we close this now?

grahamperrin:
By default, I get:  

```text
zbook/poudriere                   compression  on              local
zbook/poudriere/data              compression  on              inherited from zbook/poudriere
zbook/poudriere/data/.m           compression  on              inherited from zbook/poudriere
zbook/poudriere/data/cache        compression  off             local
zbook/poudriere/data/images       compression  on              inherited from zbook/poudriere
zbook/poudriere/data/logs         compression  on              inherited from zbook/poudriere
zbook/poudriere/data/packages     compression  off             local
zbook/poudriere/data/wrkdirs      compression  off             local
zbook/poudriere/jails             compression  on              inherited from zbook/poudriere
zbook/poudriere/jails/main        compression  on              local
zbook/poudriere/jails/main@clean  compression  -               -
zbook/poudriere/ports             compression  on              inherited from zbook/poudriere
zbook/poudriere/ports/default     compression  on              local
```

In practice (without suggesting these as defaults), I prefer: 

```text
% zfs get compression | grep poudriere
copperbowl/poudriere                                         compression  zstd-19         local
copperbowl/poudriere/data                                    compression  zstd-19         inherited from copperbowl/poudriere
copperbowl/poudriere/data/.m                                 compression  zstd-19         inherited from copperbowl/poudriere
copperbowl/poudriere/data/cache                              compression  zstd-19         inherited from copperbowl/poudriere
copperbowl/poudriere/data/logs                               compression  zstd-19         inherited from copperbowl/poudriere
copperbowl/poudriere/data/packages                           compression  zstd-19         inherited from copperbowl/poudriere
copperbowl/poudriere/data/wrkdirs                            compression  zstd-19         inherited from copperbowl/poudriere
[...]

PR #814, opened 2020-12-27
Don't hardcode many properties and try to inherit as much as possible

Duffyx:
This should fix issue #794.

I just need to get some insight if the createfs is used for anything else besides creating jails or port trees. As the mountpoint is not needed for ZFS it will just be inherited from the top dataset. Not sure if I'm missing something.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -1430,9 +1430,11 @@ get_data_dir() {
 		# Set properties on top dataset and let underlying ones inherit them
 		# Explicitly set properties for values diverging from top dataset
 		zfs create -p -o atime=off \
-			-o compression=lz4 \ 
+			-o compression=on \ 
 			-o mountpoint=${BASEFS} \
 			${ZPOOL}${ZROOTFS}
+		zfs create ${ZPOOL}${ZROOTFS}/jails
+		zfs create ${ZPOOL}${ZROOTFS}/ports
 		zfs create -o ${NS}:type=data ${ZPOOL}${ZROOTFS}/data
 		zfs create ${ZPOOL}${ZROOTFS}/data/.m
 		zfs create -o compression=off ${ZPOOL}${ZROOTFS}/data/cache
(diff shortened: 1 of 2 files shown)


Duffyx (commented):
We could also just get rid of the -p, if we create the jail dataset in common.sh.

src/share/poudriere/include/fs.sh
```diff
@@ -34,10 +34,7 @@ createfs() {
 
 	if [ -n "${fs}" -a "${fs}" != "none" ]; then
 		msg_n "Creating ${name} fs at ${mnt}..."
-		if ! zfs create -p \
-			-o compression=lz4 \
-			-o atime=off \
-			-o mountpoint=${mnt} ${fs}; then
+		if ! zfs create -p ${fs}; then
```
bdrewery, line 37: Yeah let's drop the `-p` here. If it fails trying to create outside of the ZROOTFS then something else is broken.

src/share/poudriere/common.sh
```diff
@@ -1427,14 +1427,17 @@ get_data_dir() {
...
 			return
 		fi
-		zfs create -p -o ${NS}:type=data \
-			-o atime=off \
-			-o mountpoint=${BASEFS}/data \
-			${ZPOOL}${ZROOTFS}/data
+		# Set properties on top dataset and let underlying ones inherit them
+		# Explicitly set properties for values diverging from top dataset
+		zfs create -p -o atime=off \
+			-o compression=lz4 \
```
bdrewery, line 1433: I think this should just be `compression=on`.

>       Setting compression to on indicates that the current default compression algorithm should be used.  The default balances compression and decompression speed, with compression ratio and is expected to work well on a wide variety of workloads.  Unlike all other settings for this property, on does not select a
>       fixed compression type.  As new compression algorithms are added to ZFS and enabled on a pool, the default compression algorithm may change.  The current default compression algorithm is either lzjb or, if the lz4_compress feature is enabled, lz4.

> if the lz4_compress feature is enabled, lz4.

That is also more in spirit of #794.

bdrewery (changes requested):

bdrewery (approved):

bdrewery:
Thanks!

PR #815, opened 2020-12-29
Fix git checkout during jail update to operate within SRC_BASE.

tomalexander:
The `git checkout` command during the jail update is missing the `-C ${SRC_BASE}` flag to make it operate inside the git directory. This leads to this error:

```
$ sudo poudriere jail -u -j current -t 2333220ae1a71ce51429b9d206a765df0562c987
[00:00:00] Upgrading using git+https
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```


rene0:
I created the exact same fix today :)

igalic (approved):
👍

PR #817, opened 2021-01-05
common.sh: use new URL for base system

delphij:


src/share/poudriere/common.sh
```diff
@@ -7851,7 +7851,7 @@ if [ -z "${NO_ZFS}" ]; then
 fi
 
 : ${SVN_HOST="svn.freebsd.org"}
-: ${GIT_BASEURL="github.com/freebsd/freebsd.git"}
+: ${GIT_BASEURL="github.com/freebsd/freebsd-src.git"}
```
lwhsu, line 7854: should we just use git.freebsd.org/src.git?

lwhsu:
Same with #813 ?

src/share/poudriere/common.sh
swills: Yeah, that seems like the better idea to me.

igalic: so, the problem with just switching to `git.freebsd.org/src.git` as i suggest in #813, is that this now no longer works out of the box.

on `git.freebsd.org` we don't support the `git://` protocol.

I would suggest to deprecate the `git://` protocol.
heck, i would suggest we deprecate `git+http://` too!

lwhsu: if we want to minimize the dependency, I suggest using `ssh://anongit@git.freebsd.org/src.git`.

igalic: the trouble with that is that it's very non-nonstandatd. every other mirror will provide readonly access only thru https.

uqs: Is there any reason not to use https://git.freebsd.org/src.git  (and soon ports.git). Is the fear that people run with a minimal devel/git port that can't speak TLS?

I would say that is a non-supported config and move on. It's 2021 after all, everything needs to speak https.

bdrewery: I'm on using https too. It is far more likely port 443 is open rather than 22 or 9418. Somehow getting thousands of systems using an "anonymous" ssh connection concerns me too.

And yeah github has a much better cdn than we do, so I would lean on using that. But it probably makes sense to have the canonical repo here ultimately. Now I'm wondering how github is updated and if that has risk of going stale.

Rather than get too hung up in the weeds on this I'm going to merge so it at least works without the github redirect.

Issue #459, opened 2017-04-04
PKG_NOCOMPRESS not supported


bdrewery:
```
<   Avengence>| [00:53:22] ====>> Stopping 5 builders              
<   Avengence>| [00:53:24] ====>> Creating pkgng repository        
<   Avengence>| [00:53:24] ====>> Error: Unable to extract pkg.    
```

These patches fix some of the problems but are not a complete solution:

```
diff --git src/share/poudriere/common.sh src/share/poudriere/common.sh
index ae3ad018..579b3b9f 100755
--- src/share/poudriere/common.sh
+++ src/share/poudriere/common.sh
@@ -4021,9 +4021,16 @@ ensure_pkg_installed() {
 		cp -f /usr/local/sbin/pkg-static "${mnt}/.p/pkg-static"
 		return 0
 	fi
-	[ -e ${MASTERMNT}/packages/Latest/pkg.txz ] || return 1 #pkg missing
-	injail tar xf /packages/Latest/pkg.txz -C / \
-		-s ",/.*/,.p/,g" "*/pkg-static"
+	if [ -e "${MASTERMNT}/packages/Latest/pkg.txz" ]; then
+		injail tar xf /packages/Latest/pkg.txz -C / \
+		    -s ",/.*/,.p/,g" "*/pkg-static"
+	elif [ -e "${MASTERMNT}/packages/Latest/pkg.tar" ]; then
+		injail tar xf /packages/Latest/pkg.tar -C / \
+		    -s ",/.*/,.p/,g" "*/pkg-static"
+	else
+		# pkg missing
+		return 1
+	fi
 	return 0
 }
 
@@ -5480,6 +5487,12 @@ build_repo() {
 		elif [ -n "${PKG_REPO_SIGNING_KEY}" ]; then
 			sign_pkg pubkey "${PACKAGES}/Latest/pkg.txz"
 		fi
+	elif [ -e "${PACKAGES}/Latest/pkg.tar" ]; then
+		if [ -n "${SIGNING_COMMAND}" ]; then
+			sign_pkg fingerprint "${PACKAGES}/Latest/pkg.tar"
+		elif [ -n "${PKG_REPO_SIGNING_KEY}" ]; then
+			sign_pkg pubkey "${PACKAGES}/Latest/pkg.tar"
+		fi
 	fi
 }
 ```

```
diff --git src/share/poudriere/common.sh src/share/poudriere/common.sh
index ae3ad018..a3a34e0d 100755
[...]

PR #825, opened 2021-01-19
Reuse WRKDIR_ARCHIVE_FORMAT for the value of PKG_FORMAT in pkgbase

igalic:
pass on WRKDIR_ARCHIVE_FORMAT as PKG_FORMAT to pkgbase package builds


bdrewery:
I think PKG_EXT/PKG_SUFX is more proper here and relates to #459 somewhat. The extensions are all hardcoded currently.
WRKDIR_ARCHIVE_FORMAT is for debug tarballs of work directories and not the package format.

igalic:
in that case we should maybe call it `PKG_FORMAT` :D

PR #830, opened 2021-02-02
git+ssh: the git URL based on ssh starts with git@

pizzamig:
`ssh` access to git repository is not using URL based on `ssh://` but with `git@github.com:name/repo` or `git@gitlab.com:name/repo`

This is true for github and gitlab.
This is needed to support git private repositories.

### DIFF
--- a/src/share/poudriere/ports.sh
+++ b/src/share/poudriere/ports.sh
@@ -153,7 +153,7 @@ if [ -n "${SOURCES_URL}" ]; then
 		;;
 	git*)
 		case "${SOURCES_URL}" in
-		ssh://*) METHOD="git+ssh" ;;
+		git@*) METHOD="git+ssh" ;;
 		http://*) METHOD="git+http" ;;
 		https://*) METHOD="git+https" ;;
 		git://*) METHOD="git" ;;
--- a/src/share/poudriere/ports.sh
+++ b/src/share/poudriere/ports.sh
@@ -154,6 +154,7 @@ if [ -n "${SOURCES_URL}" ]; then
 	git*)
 		case "${SOURCES_URL}" in
 		git@*) METHOD="git+ssh" ;;
+		ssh://*) METHOD="git+ssh" ;;
 		http://*) METHOD="git+http" ;;
 		https://*) METHOD="git+https" ;;
 		git://*) METHOD="git" ;;
--- a/src/share/poudriere/ports.sh
+++ b/src/share/poudriere/ports.sh
@@ -153,12 +153,13 @@ if [ -n "${SOURCES_URL}" ]; then
 		;;
 	git*)
 		case "${SOURCES_URL}" in
-		git@*) METHOD="git+ssh" ;;
 		ssh://*) METHOD="git+ssh" ;;
 		http://*) METHOD="git+http" ;;
 		https://*) METHOD="git+https" ;;
 		git://*) METHOD="git" ;;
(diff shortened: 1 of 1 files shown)


igalic (approved):

kevans91:
> `ssh` access to git repository is not using URL based on `ssh://` but with `git@github.com:name/repo` or `git@gitlab.com:name/repo`
> 

The issue is your formatting; git just uses the format that the underlying ssh does. The correct invocation you're looking for is `-m git -U ssh://git@gitlab.com/name/repo` -- mainly, note the : -> / swap. Hardcoding the git@ user at something to identify is likely not what we want to be doing here, but documenting the difference in how people use git vs. what poudriere wants (valid URI).

pizzamig:
Hi Kyle, thanks for the answer.
I know that `poudriere` would accept a different formatting.
The `git@` is not MY formatting, it's what Github and Gitlab provides and `git` supports it, while `poudriere` is the only tool I know where I need to change it.
Yes, you can improve the documentation, but from a user prospective, I miss the point to not just accept this very popular and vastly adopted formatting

kevans91:
The main thing that I take exception with here is the hardcoded user. This should instead match a generic *:* type pattern and assume it's ssh in the face of no other spec, because there's many times where the server side user will not be git and times when it will be omitted entirely. The minimum you need is a host name and a path, and you can't assume the format of either.

kevans91:
You should probably match \*://\* before it so you can complain about an unknown protocol.

pizzamig:
The user is hardcoded and this is the way `ssh` works for Github and Gitlab.
This patch has the only purpose to recognize the format provided by Github and Gitlab for `ssh`, not to replace the current `ssh://` URL format, that I agree is preferable, but not provided by those 2 popular providers.
For the poudriere repo, Github provides this: `git@github.com:freebsd/poudriere.git`
If you want to restrict the purpose, I can refine the match to `git@*:*` , but there is no`://` to match.
We can even go to `git@git*:*` to be more specific on the intention to support Github and Gitlab and not other generic cases

kevans91:
> The user is hardcoded and this is the way `ssh` works for Github and Gitlab.
> This patch has the only purpose to recognize the format provided by Github and Gitlab for `ssh`, not to replace the current `ssh://` URL format, that I agree is preferable, but not provided by those 2 popular providers.
> For the poudriere repo, Github provides this: `git@github.com:freebsd/poudriere.git`
> If you want to restrict the purpose, I can refine the match to `git@*:*` , but there is no`://` to match.
> We can even go to `git@git*:*` to be more specific on the intention to support Github and Gitlab and not other generic cases

I think you're misunderstanding what I'm saying. This is specifically what I'm recommending in lieu of this patch and what I would advocate for; tack these two cases on to the end, between file:// and the catch-all:

```
*://*) err 1 "Invalid git protocol" ;;
*:*) METHOD="git+ssh" ;;
```

So we can catch absolutely bogus things (protocol) while still passing the ssh-looking specification through to ssh to let it reject it instead if it wants to. This solves the problem not just for you, but also for me and everyone else with different circumstances.

pizzamig:
Yes, I misunderstood what you were saying (the 'before' part).
I'll rework the patch as soon as possible.
Thanks

kevans91 (approved):
LGTM, thanks! :-) IMO we can refine the new case as needed if we find a scenario where it's just flat wrong to let ssh try it, but I suspect we won't find one in any remotely conventional usage.

PR #833, opened 2021-02-16
Don't allow /dev/bpf* in jails

AMDmi3:
net/libnet10 which bpf device was enabled for was removed from the ports tree, net/libnet does not need this and I've just fixed net/libdnet.
If any port ever needs this, the port must be fixed instead - ports must never need to access bpf during the build.

This effectively reverts e3a299603ece5b27da22284d58e2dd6afead30c05

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -1664,7 +1664,7 @@ do_jail_mounts() {
 	local from="$1"
 	local mnt="$2"
 	local name="$3"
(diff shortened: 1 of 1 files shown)


bdrewery:
```
~/git/poudriere6 # git show e3a299603ece5b27da22284d58e2dd6afead30c05
fatal: ambiguous argument 'e3a299603ece5b27da22284d58e2dd6afead30c05': unknown revision or path not in the working tree.
```

Can you double check that commit please and update the commit message? Then it's good to go.

PR #845, opened 2021-03-15
Prepare older installs to take advantage of inheritance

### LINKED ISSUE #844: poudriere-devel: Jail creation fails on older installs
After 1b3cf3a  There are problems with this for older installs.. I have:

```
ZPOOL=scratch
```

In poudriere.conf and my jails dataset has the following mountpoint:

```
NAME                     PROPERTY    VALUE                     SOURCE
scratch/poudriere/jails  mountpoint  /scratch/poudriere/jails  default
```

So it creates a dataset without a mount point and then cannot find it:

```
# poudriere -vv jail -c -j 122 -v 12.2-RELEASE
[00:00:00] Creating 122 fs at /usr/local/poudriere/jails/122... done
mkdir: /usr/local/poudriere/jails/122: No such file or directory
```

If I update the mountpoint property for `scratch/poudriere/jails` it works fine, but there needs to be an upgrade path.

@driesmp Do you have any thoughts on how this should be handled?

driesmp:
With new way of ZFS dataset creation, going from an explicit mountpoint to an inherited mountpoint there was a problem with existing dataset structures not having the correct mountpoint assigned resulting in a failed jail or port tree creation. This fixes #844.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -1414,6 +1414,13 @@ get_data_dir() {
 	fi
 
 	if [ -z "${NO_ZFS}" ]; then
+		# Prepare older installs to take advantage of inheritance
+		zrootfsmountpoint=$(zfs get -o value -H mountpoint ${ZPOOL}${ZROOTFS})
+		if [ "${zrootfsmountpoint}" != ${BASEFS} ]; then
+			zfs inherit -r mountpoint ${ZPOOL}${ZROOTFS}
+			zfs set mountpoint=${BASEFS} ${ZPOOL}${ZROOTFS}
+		fi
+		# Normal place to look for the data mountpoint
 		data=$(zfs list -rt filesystem -H -o ${NS}:type,mountpoint ${ZPOOL}${ZROOTFS} 2>/dev/null |
 		    awk '$1 == "data" { print $2; exit; }')
 		if [ -n "${data}" ]; then
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -1415,7 +1415,7 @@ get_data_dir() {
 
 	if [ -z "${NO_ZFS}" ]; then
 		# Prepare older installs to take advantage of inheritance
-		zrootfsmountpoint=$(zfs get -o value -H mountpoint ${ZPOOL}${ZROOTFS})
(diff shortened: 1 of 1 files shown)


src/share/poudriere/common.sh
```diff
@@ -1414,6 +1414,13 @@ get_data_dir() {
 	fi
 
 	if [ -z "${NO_ZFS}" ]; then
+		# Prepare older installs to take advantage of inheritance
+		zrootfsmountpoint=$(zfs get -H -o value mountpoint ${ZPOOL}${ZROOTFS})
+		if [ "${zrootfsmountpoint}" != ${BASEFS} ]; then
+			zfs inherit -r mountpoint ${ZPOOL}${ZROOTFS}
+			zfs set mountpoint=${BASEFS} ${ZPOOL}${ZROOTFS}
```
bdrewery, line 1421: Please quote all the strings.

src/share/poudriere/common.sh
```diff
@@ -1414,6 +1414,13 @@ get_data_dir() {
 	fi
 
 	if [ -z "${NO_ZFS}" ]; then
+		# Prepare older installs to take advantage of inheritance
+		zrootfsmountpoint=$(zfs get -H -o value mountpoint ${ZPOOL}${ZROOTFS})
+		if [ "${zrootfsmountpoint}" != ${BASEFS} ]; then
+			zfs inherit -r mountpoint ${ZPOOL}${ZROOTFS}
```
bdrewery, line 1420: I don't like the idea of forcing this change recursively on everyone. It's one thing to have the mountpoint forced at creation time but the user may have changed it for some reason since then.

src/share/poudriere/common.sh
```diff
@@ -1414,6 +1414,13 @@ get_data_dir() {
 	fi
 
 	if [ -z "${NO_ZFS}" ]; then
+		# Prepare older installs to take advantage of inheritance
```
bdrewery, line 1417: We still have the problem of /jails not getting fixed for `atime=off` and `compression=lz4` removed in https://github.com/freebsd/poudriere/pull/814/files

bdrewery: See update in https://github.com/freebsd/poudriere/commit/1c9357d08e35f2bb248533ce043b2e61920ebf9e

PR #848, opened 2021-03-27
jail: unbreak git checkout of a local clone

jbeich:
```
$ fgrep GIT /usr/local/etc/poudriere.conf
GIT_BASEURL=/path/to/freebsd-src

$ git clone https://git.freebsd.org/src.git /path/to/freebsd-src
$ poudriere jail -cj main-amd64 -a amd64 -v main -m git+file
[00:00:00] Creating main-amd64 fs at /poudriere/jails/main-amd64... done
[00:00:00] Checking out the sources with git+file...fatal: protocol '' is not supported
[00:00:00] Error:  fail
[00:00:00] Error while creating jail, cleaning up.
[00:00:00] Removing main-amd64 jail... done
[00:00:00] Cleaning main-amd64 data... done
```

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -1254,8 +1254,8 @@ if [ -n "${SOURCES_URL}" ]; then
 		ssh://*) METHOD="git+ssh" ;;
 		http://*) METHOD="git+http" ;;
 		https://*) METHOD="git+https" ;;
+		file://*) METHOD="git+file" ;;
 		git://*) METHOD="git" ;;
-		file://*) METHOD="git" ;;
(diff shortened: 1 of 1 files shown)


bdrewery:
How do we unbreak something that isn't documented or codified? `git+file` hasn't been a thing from what I see.
And what about `poudriere-ports`?
Manpages?
Usage output?

Not directed at you, but this is getting out of hand with the differences in git protocol handling between ports/jail. Every few weeks there's a new fix and a new regression.

mat813:
I think anything that is git* should accept an "argument" and be given to git verbatim and let git clone/fetch/pull handle it, no need to support all and every quirks that it can do.

PR #851, opened 2021-04-08
Update poudriere-ports manpage

DavidMarec:
Add a short description, taken from the _usage_ output, about the "-U" flag.


bdrewery:
Thank you for doing this. I would merge it but there's significantly more work needed to update all of the places that need to document this. I'll link the commit to this PR when it comes in.

PR #853, opened 2021-04-13
change Makefile identification for logfiles

linimon:
With the switch to git, the logfile line starting with "Makefile ident" has become useless (ident(1) returns empty, as expected).

To help people figure out which version of the port was run, switch it out for an ls -l of the Makefile itself.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -844,7 +844,7 @@ buildlog_start() {
 	echo "package name: ${pkgname}"
 	echo "building for: $(injail uname -a)"
 	echo "maintained by: ${mk_MAINTAINER}"
-	echo "Makefile ident: $(ident -q ${mnt}/${portdir}/Makefile|sed -n '2,2p')"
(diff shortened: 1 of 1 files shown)


bdrewery:
I like the idea but I believe that the datestamp here will be when it was checked out rather than the git commit date.

Maybe this could work? It would try to find the last commit hash or just fallback to the `ls` if `git` doesn't work.
```
	echo "Makefile datestamp: $(${GIT_CMD} -C "${mnt}/usr/ports" log -n1 --format=%H "${mnt}/${portdir}/Makefile" 2>/dev/null || ls -l ${mnt}/${portdir}/Makefile)"
```

linimon:
I am completely indifferent to what's in it as long as it is nonempty :-)

As git and I are not yet friends, I cannot comment on whether it would work or not.  I suppose I can feed it to ppcdevref and take a look.

bdrewery:
Hm well with git shallow cloning this won't work either.

linimon:
And the comment isn't quite right.  From ppcdevref:

Makefile datestamp: 52dab67dd2f28332f96a34e81e23d3f29ae5ffa5

Issue #583, opened 2018-01-20
Options command forces custom options directory to be associated with a ports tree


davehayes:
From the man page:

```
 Before building a package, poudriere can mount a directory containing option files if available. poudriere will check for any of these directories in this order:
    /usr/local/etc/poudriere.d/<jailname>-<tree>-<setname>-options
    /usr/local/etc/poudriere.d/<jailname>-<setname>-options
    /usr/local/etc/poudriere.d/<jailname>-<tree>-options
    /usr/local/etc/poudriere.d/<setname>-options
    /usr/local/etc/poudriere.d/<tree>-options
    /usr/local/etc/poudriere.d/<jailname>-options
    /usr/local/etc/poudriere.d/options

 If a directory with this name exists, it is null-mounted into the
 /var/db/ports/ directory of the jail, thus allowing to build package with
 custom OPTIONS.

 The options subcommand can be used to easily configure options in the correct directory.
```
Apparently, for the options subcommand, "the correct directory" is not
flexible, like implied above. Unless I use the 'default' ports tree, I can not
(easily) configure options in any of the following directories:

```
    /usr/local/etc/poudriere.d/<jailname>-<setname>-options
    /usr/local/etc/poudriere.d/<setname>-options
    /usr/local/etc/poudriere.d/<jailname>-options
    /usr/local/etc/poudriere.d/options
```
This is because the -p switch forces 'options' to use one of

```
    /usr/local/etc/poudriere.d/<jailname>-<tree>-<setname>-options
    /usr/local/etc/poudriere.d/<jailname>-<tree>-options
    /usr/local/etc/poudriere.d/<tree>-options
```
Of course, the behavior I'd expect to see is the same behavior as make.conf or
poudriere.conf; an 'overlay' of options directories with the most specific
options dir being used. This is because of the customizable behavior of
make.conf, src.conf, and poudriere.conf.

Here's how I discovered this. When I upgraded to 3.2.4, I got this dialog:

```
  > poudriere options -f myportlist -j myoldjail -p MYPORTSNAPSHOT
 [00:00:00] Warning: You already have options configured without '-p MYPORTSNAPSHOT' that will no longer be used.
 [00:00:00] Warning: Drop the '-p MYPORTSNAPSHOT' option to avoid this problem.
 [00:00:00] Are you sure you want to continue? [y/N] ^C
```

This was really surprising. Looking at the code I see this:

 ```sh
[...]

PR #861, opened 2021-04-26
Sort options and document -p flag

0mp:


src/bin/poudriere-options.8
```diff
@@ -62,25 +62,25 @@ This command accepts the following options:
...
 .Ar category/port
 and shell-style comments are allowed.
 .It Fl j Ar jailname
 If given, configure the options only for the given jail.
-.It Fl p Ar portstree
-Run the configuration inside the given ports tree (by default uses default)
 .It Fl n
 Do not be recursive
+.It Fl p Ar portstree
+Run the configuration inside the given ports tree.
```
bdrewery, line 82: Might be worth mentioning that any of `-z`, `-p`, and `-j`, used will cause the options to be saved to a specific directory for those flags.
#583 covers the issue with "using" versus "binding" for each of these. Currently they bind.

bdrewery (approved):

PR #887, opened 2021-07-12
Add new configuration variable BUILD_JAIL_SILENTLY

yasuhirokimura:
Add new configuration variable `BUILD_JAIL_SILENTLY`. If you set it to `yes` in poudriere.conf, then `-s` option is passed to make(1) and therefore less messages are displayed when building/installing/updating jail from source tree. It affects only when you use either `-b` or `-m git|svn` option of `jail` command.

### DIFF
--- a/src/etc/poudriere.conf.sample
+++ b/src/etc/poudriere.conf.sample
@@ -322,6 +322,12 @@ DISTFILES_CACHE=/usr/ports/distfiles
 # Default: no
 #HTML_TRACK_REMAINING=yes
 
+# Display less messages when building/installing/updating jail from
+# source tree, Affect only when you use either '-b' or '-m git|svn.'
+# option of `jail` command,
+# Default: no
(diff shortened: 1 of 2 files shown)


bdrewery:
I think we have too many options as is and complicate the interface. Defaulting to `make -s` when `poudriere -v` is not used seems better to me. Git/svn already do that.
Somewhere near the top:
```
[ $VERBOSE -gt 0 ] && MAKE_CMD="${MAKE_CMD} -s"
```

evadot:
I have MAKEWORLDARGS=" -s " in my poudriere.conf for years now so any solution would be fine to me.

yasuhirokimura:
> I think we have too many options as is and complicate the interface. Defaulting to `make -s` when `poudriere -v` is not used seems better to me. 

I updated that way.

bdrewery:
Thank you for your patience in getting this merged, and keeping up with rebasing!

PR #892, opened 2021-07-25
Allow to specify base mount point for jails using JAIL_BASEMNT

jurajlutter:
It is sometimes desirable (for example, when poudriere data is on NFS, as in my case) to have base jail mount on local filesystem to avoid unnecessary NFS traffic.

This patch adds `JAIL_BASEMNT` variable that can override the default `${POUDRIERE_DATA}/.m`


bdrewery:
This is probably fine but my head isn't in the right place to do a thorough review right now.

PR #912, opened 2021-08-29
Allow to specify jail base mount directory.

jurajlutter:
It is sometimes desirable (for example, when poudriere data is on NFS, as in my case) to have base jail mount on local filesystem to avoid unnecessary NFS traffic.

This patch adds `JAIL_BASEMNT` variable that can override the default `${POUDRIERE_DATA}/.m`

### DIFF
--- a/src/poudriered/poudriered.c
+++ b/src/poudriered/poudriered.c
@@ -897,6 +897,10 @@ main(int argc, char **argv)
 
 	const ucl_object_t *sock_path_o, *pidfile_path_o, *foreground_o;
 
+	if (argc == 2 && strcmp(argv[1], "-f") == 0) {
+		foreground = true;
+	}
+
 	if (argc == 3) {
 		struct ucl_parser *parser = NULL;
 


src/share/poudriere/common.sh
```diff
@@ -8328,6 +8337,10 @@ POUDRIERE_DATA="$(get_data_dir)"
 if [ -e "${POUDRIERE_DATA}" ]; then
 	POUDRIERE_DATA=$(realpath "${POUDRIERE_DATA}")
 fi
+JAIL_BASEMNT="$(get_jail_base_mnt)"
```
bdrewery, line 8340: `get_jail_base_mnt` is not needed. The following will provide a default and use whatever a user overrides with.
```
: ${JAIL_BASEMNT:=${POUDRIERE_DATA}/.m}
```

src/etc/poudriere.conf.sample
```diff
@@ -354,3 +354,7 @@ DISTFILES_CACHE=/usr/ports/distfiles
 # be fetched.
 # Default: everything
 #PACKAGE_FETCH_WHITELIST="gcc* rust llvm*"
+
+# Set the base mount point for jails
+# Default:
+#JAIL_BASEMNT=${POUDRIERE_DATA}/.m
```
bdrewery, line 360: I think we should not put this here as it is not something a user typically needs to modify, and the default is confusing and error-prone since POUDRIERE_DATA can be dynamic / undefined here.

bdrewery (changes requested):

src/etc/poudriere.conf.sample
jurajlutter: Where should this belong to?

src/share/poudriere/common.sh
jurajlutter: Fixed.

bdrewery:
Somehow this branch was rebased to include another commit modifying logclean.

PR #921, opened 2021-09-17
Fix booting ZFS images.

### LINKED ISSUE #735: image: zrawdisk image does not boot
Hi,
An image built using `zrawdisk` type doesn't boot as support for non-partitioned root ZFS disks has been dropped by https://svnweb.freebsd.org/base?view=revision&revision=342151

davidchisnall:
FreeBSD now requires ZFS bootable vdevs to live in partitions, rather
than be raw disks.  This change creates a GPT partition table and
installs both MBR and UEFI boot loaders.

This change also uses the image name as a *temporary* name for the zpool
so that zroot can be consistently called zroot, without introducing a
conflict with the host system's zroot.  If there is a use case for
having a non-default name for zroot then this can be changed.

Tested booting an image as a Hyper-V Gen1 (MBR) and Gen2 (UEFI) VM.

zrawdisk is no longer a raw disk and so probably needs renaming.

Fixes #735


davidchisnall:
It sounds as if some folks were using the raw image support with `mkimg`.  If this is useful to other people, I can factor out the common bits between ZFS raw image and ZFS + partition table and provide both.

evadot:
> It sounds as if some folks were using the raw image support with `mkimg`. If this is useful to other people, I can factor out the common bits between ZFS raw image and ZFS + partition table and provide both.

That's the goal of the zrawdisk type, if you want some partition table I suggest you put that in a new zdisk type image.

davidchisnall:
I'll try to refactor this to pull out the common bits and add a bootable zdisk type either this week or at the weekend.

davidchisnall:
@evadot, I've done the refactoring but I can't figure out how to install the image_helpers.sh file.  Everything I worked on that used GNU autotools moved to CMake a decade ago and so I've never had to touch a build system like this before.

davidchisnall:
I have tried:

 - Adding it to Makefile.am (doesn't seem to regenerate the things that are needed)
 - Rerunning ./configure (Doesn't seem to do anything)
 - Running ./autogen.sh (Generates a configure that contains a syntax error)

I am now giving up.

davidchisnall:
All checks passing, I believe this should be ready to land.

evadot:
> All checks passing, I believe this should be ready to land.

Well commit shouldn't say "Fix booting ZFS images." as this doesn't fix anything, that adds a new type.
Other than that I really don't like having the zdisk image in the same file as the rawdisks one, this really should be a different file otherwise things will start to be unreadable at one point (as it was before).

evadot:
> > Well commit shouldn't say "Fix booting ZFS images." as this doesn't fix anything, that adds a new type.
> 
> Welcome to GitHub, where you can edit commit messages when you press the squash and merge button.

 I won't do that.

> > Other than that I really don't like having the zdisk image in the same file as the rawdisks one, this really should be a different file otherwise things will start to be unreadable at one point (as it was before).
> 
> This patch is needed for me to be able to build Azure-bootable FreeBSD disk images. I've been trying to get it merged for over a month now. Please either:
> 
>     * Review it and give me a complete list of changes that you require (if this requires adding new files, please just push the change directly - I can't make this project's '80s build system do anything that it doesn't already do)

 Move the zdisk image stuff in image_zdisk.sh and no I won't do the change myself, you're a grown man.

>     * Merge it as-is and make any improvements that you want afterwards, or
> 
>     * Close it and the original issue as WONTFIX so I can give up on Poudriere and work on something else.

PR #936, opened 2021-10-31
image: use parameter substitution to parse $MEDIATYPE

allanjude:
Replace `$(echo ${MEDIATYPE} | cut -d '+' -f 1)` with shell builtins

### DIFF
--- a/src/share/poudriere/image.sh
+++ b/src/share/poudriere/image.sh
@@ -388,8 +388,10 @@ post_getopts
 : ${IMAGENAME:=poudriereimage}
 MASTERNAME=${JAILNAME}-${PTNAME}${SETNAME:+-${SETNAME}}
 
-MAINMEDIATYPE=$(echo ${MEDIATYPE} | cut -d '+' -f 1)
-SUBMEDIATYPE=$(echo ${MEDIATYPE} | cut -d '+' -f 2)
+MAINMEDIATYPE=${MEDIATYPE%%+*}
+MEDIAREMAINDER=${MEDIATYPE#*+}
+SUBMEDIATYPE=${MEDIAREMAINDER%%+*}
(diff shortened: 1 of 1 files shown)


src/share/poudriere/image.sh
```diff
@@ -388,8 +388,10 @@ post_getopts
 : ${IMAGENAME:=poudriereimage}
 MASTERNAME=${JAILNAME}-${PTNAME}${SETNAME:+-${SETNAME}}
 
-MAINMEDIATYPE=$(echo ${MEDIATYPE} | cut -d '+' -f 1)
-SUBMEDIATYPE=$(echo ${MEDIATYPE} | cut -d '+' -f 2)
+MAINMEDIATYPE=${MEDIATYPE%%+*}
+MEDIAREMAINDER=${MEDIATYPE#*+}
```
evadot, line 392: Are the REMAINDER useful ?

allanjude: Yes. It is used for stuff like: zfs+send+be vs zfs+send+full

evadot: Right but the variable aren't used now, so either use them or remove them.

evadot (changes requested):

src/share/poudriere/image.sh
allanjude: Using them is added in the other pull request. I tried to separate the logical changes. If you really want, I can move the remainder bit to being added in the other pull request.

https://github.com/freebsd/poudriere/pull/731/files#diff-968baa06df11de3b0484d4ca355915b96462d79dc0a5d659b4c834f91c56a6d6R131

evadot: Ah ok I missed that. Thanks.

PR #947, opened 2021-11-12
image_zfs: Fix inverted logic

jlduran:
`zpool list -Ho ${ZFS_POOL_NAME}` exits with `0` if the name exists. This prevents overwriting the existing pool.

Fixes: ae641ed4ff326fb338aff781506e4b06ead65bfc

cc/ @allanjude


allanjude:
Good catch, thank you.

@evadot Can we get this merged quickly as a followup fix

jlduran:
@allanjude I have another fix (that avoids a segfault), I haven’t fully investigated the cause, but will sumbit a PR just in case. Thank you!

PR #948, opened 2021-11-12
image_zfs: Do not force a size

jlduran:
It is segfaulting.  Further investigation is needed.  Remove for now.

cc/ @allanjude


davidchisnall:
This doesn't segfault for me, it just raises an error.  The problem is that the :512K is in the wrong place.  It should be:

```
            gptboot="-p freebsd-boot:512K=${mnt}/boot/gptzfsboot"
```

jlduran:
Followup #962

PR #952, opened 2021-11-17
Fix thin jail creation

brd:
Some jails are very thin and do not have /usr/include/sys populated,
which leads to this error on jail creation:

awk: can't open file /usr/local/poudriere/jails/testjail/usr/include/sys/param.h
 source line number 1

So use the param.h from the src tree instead.


evadot:
I think that there is more chance of not having src than /usr/include/sys (or at least that's where we should go)

PR #953, opened 2021-11-17
Allow to disable grepping logs for failure reason

AMDmi3:
This takes a lot of time for no usable result in most cases

### DIFF
--- a/src/etc/poudriere.conf.sample
+++ b/src/etc/poudriere.conf.sample
@@ -365,3 +365,7 @@ DISTFILES_CACHE=/usr/ports/distfiles
 # be fetched.
 # Default: everything
 #PACKAGE_FETCH_WHITELIST="gcc* rust llvm*"
+
+# Grep build logs to determine a possible build failure reason.
+# Default: yes
+#DETERMINE_BUILD_FAILURE_REASON=yes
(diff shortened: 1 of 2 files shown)


bdrewery:
I think this is fine but I'm curious if you have seen measurable improvements with disabling this?

swills:
For me, it can take literally hours to run processonelog.sh on some larger log files, so I would really like this change.

AMDmi3:
I occasionally run into ports which generate huge logs, and grepping after a failure may take a few dozen of seconds on them. This is terribly annoying for change→test→repeat workflow (and I don't even see the result after testport). I don't remember specific ports to demonstrate this, unfortunately.

bdrewery:
> This is terribly annoying for change→test→repeat workflow (and I don't even see the result after testport).

That's an excellent point. This could be moved off to an asynchronous job too. I'll merge in what you have for now.

AMDmi3:
>  I'll merge in what you have for now.

I'd :+1: that, but you seem to be blocking me for some reason.

clausecker:
Same issue here.  When a larger port (e.g. `qt5-webkit`) fails due to an OOM condition, a 60M+ log file is created.  Analysing the error cause takes more than an hour due to `grep` calls that take more than 20 minutes each.  I'm running Poudriere on a Raspberry Pi 4B+ running arm64 FreeBSD 13.

PR #963, opened 2021-12-07
Improvements to poudriere-bulk.8

0mp:
A general clean up and some wordsmithing.

I'm not sure about this part though:

```
With this option
.Nm poudriere
will first try to fetch a binary package
from the specified binary package repository
and only then check if the package already exist.
```

What do we want to tell the user about here?

### DIFF
--- a/src/man/poudriere-bulk.8
+++ b/src/man/poudriere-bulk.8
@@ -28,7 +28,7 @@
 .\"
 .\" Note: The date here should be updated whenever a non-trivial
 .\" change is made to the manual page.
-.Dd April 29, 2022
+.Dd June 29, 2022
 .Dt POUDRIERE-BULK 8
 .Os
 .Sh NAME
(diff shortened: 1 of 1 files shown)


bdrewery:
> With this option
.Nm poudriere
will first try to fetch a binary package
from the specified binary package repository

keep

> and only then check if the package already exist.

The part isn't needed anymore. The fetching tries to be smart enough to not do needless fetches.

src/man/poudriere-bulk.8
```diff
@@ -139,19 +139,27 @@ By default
...
 .Ar url .
-With this option poudriere will first try to fetch from the binary package
-repository specified the binary package prior to do the sanity check if the
-package does not already exist.
+.Pp
+With this option
+.Nm poudriere
+will first try to fetch a binary package
+from the specified binary package repository
+and only then check if the package already exist.
```
bdrewery, line 162: This last part can drop

bdrewery (changes requested):

0mp:
Done.

PR #974, opened 2022-01-28
Add hybrid ISO image support

allanjude:
This makes a version of .iso that can also be written to a USB stick

Sponsored-by: Zenith Electronics, LLC
Sponsored-by: Klara Inc.


allanjude:
Rebased

PR #979, opened 2022-02-17
testport: don't try to delete /compat/linux

brooksdavis:
testport clears PREFIX before building.  When PREFIX is /compat/linux
this fails because /compat/linux/proc is mounted.  As such, exclude it
like /usr is excluded.

### DIFF
--- a/src/share/poudriere/testport.sh
+++ b/src/share/poudriere/testport.sh
@@ -298,7 +298,7 @@ fi
 [ "${PREFIX}" != "${LOCALBASE}" ] && PORT_FLAGS="PREFIX=${PREFIX}"
 msg "Building with flags: ${PORT_FLAGS}"
 
-if [ -d ${MASTERMNT}${PREFIX} -a "${PREFIX}" != "/usr" ]; then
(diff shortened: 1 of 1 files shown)


bdrewery:
Perhaps `rm -Rfx` to avoid removing `/compat/linux/proc`?

arrowd:
Oh, so this is why testport doesn't work on Linuxulator ports? It'd be great to get this fix in.

arrowd:
When doing `testport` on a linux port with Poudriere 3.4.2 I still get an error

```
[00:00:02] Sanity checking build queue
[00:00:02] Processing PRIORITY_BOOST
[00:00:02] Balancing pool
[00:00:02] Recording filesystem state for prepkg... done
[00:00:03] Committing packages to repository: /usr/local/poudriere/data/packages/sbreeze-default/.real_1733987497 via .latest symlink
[00:00:03] Removing old packages
[00:00:03] Building with flags: PREFIX=/compat/linux
[00:00:03] Removing existing /compat/linux
rm: /usr/local/poudriere/data/.m/sbreeze-default/ref/compat/linux/proc: Device busy
rm: /usr/local/poudriere/data/.m/sbreeze-default/ref/compat/linux: Directory not empty
[ERROR] Unhandled error!
[00:00:03] Cleaning up
[00:00:03] Unmounting file systems
Exiting with status 1
```

PR #984, opened 2022-03-08
image_zfs: Allow same-as-host pool names

jlduran:
Use `zpool create -t` to set the in-core pool name to a temporary, random pool name, while the on-disk name will be the name specified as the pool name in `ZFS_POOL_NAME`, by default `zroot` to match what is currently used on release images.

Keep the check for a possible name collision on the randomly-created pool name, to err on the side of caution.

Inspired by:	https://reviews.freebsd.org/D34426

cc/ @allanjude

### DIFF
--- a/src/share/poudriere/image.sh
+++ b/src/share/poudriere/image.sh
@@ -395,7 +395,7 @@ post_getopts
 : ${SWAPSIZE:=0}
 : ${PTNAME:=default}
 : ${ZFS_SEND_FLAGS:=-Rec}
-: ${ZFS_POOL_NAME:=${IMAGENAME}root}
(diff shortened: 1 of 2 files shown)


allanjude:
There might be a slightly better way to generate the temporary name to more strong ensure there won't be a collision, but I think the jot -r is fine.

I like this idea.

jlduran:
> There might be a slightly better way to generate the temporary name to more strong ensure there won't be a collision, but I think the jot -r is fine.

I had the exact same thought, tried `basename $(mktemp -t ${ZFS_POOL_NAME})` and `uuidgen | tr -d '-'` (maybe also adding a timestamp: `date +%s`), but after staring at it for a while, reverted to the original `jot -r`.

PR #987, opened 2022-03-16
Fix tmpfs blacklist wildcards

overhacked:
Disabling globs for the loop that checks currently
building packages against `TMPFS_BLACKLIST` or wildcards in
`TMPFS_BLACKLIST` will expand to files in the current
directory, instead of being passed into the case statement
as a pattern.

Also fixed a breaking typo in code that cleans up from `TMPFS_BLACKLIST_TMPDIR`.


bdrewery:
Thanks for this. I didn't realize these would expand in `for`. That means we have a lot of other places that need the same fix.
`PACKAGE_FETCH_BLACKLIST`, `PRIORITY_BOOST`, `PACKAGE_FETCH_WHITELIST`, `ALLOW_NETWORKING_PACKAGES`. Can you fix those too please?

bdrewery (approved):

bdrewery:
Made #1001 for the other ones to fix.

PR #988, opened 2022-03-17
Fix setting the hostname when cross building images

jlduran:
Do not use `sysrc -R`, it tries to run a shell inside the chroot(8), if the files inside are built for a different architecture it fails:

    chroot: /bin/sh: Exec format error

As a workaround, work directly from the host to the /etc/rc.conf target file (`sysrc -f`).

### DIFF
--- a/src/share/poudriere/image.sh
+++ b/src/share/poudriere/image.sh
@@ -532,7 +532,14 @@ pwd_mkdb -d ${WRKDIR}/world/etc -p ${WRKDIR}/world/etc/master.passwd
 
 # Set hostname
 if [ -n "${HOSTNAME}" ]; then
-	sysrc -q -R "${WRKDIR}/world" hostname="${HOSTNAME}"
+	# `sysrc -R` tries to run a shell inside the chroot(8).
+	# It may fail if the target is on a different architecture than the host.
+	# In this case, set /etc/rc.conf as the destination for the hostname.
+	if [ "${arch}" == "${host_arch}" ]; then
+		sysrc -q -R "${WRKDIR}/world" hostname="${HOSTNAME}"
+	else
+		sysrc -q -f "${WRKDIR}/world/etc/rc.conf" hostname="${HOSTNAME}"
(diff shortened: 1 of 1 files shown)


jlduran:
Related to: #934

swills:
I've run into this as well, looks like a good solution

jlduran:
I'll test using:
```sh
RC_CONFS="${WRKDIR}/world/etc/rc.conf ${WRKDIR}/world/etc/rc.conf.local" \
    RC_DEFAULTS="${WRKDIR}/world/etc/defaults/rc.conf" \
    sysrc -q hostname="${HOSTNAME}"
```
instead, it should be a little better.  Or maybe I'll look into `HLINK_FILES`.

jlduran:
My testing on this subject has been halted for a while.
In the meantime, add a common check (`if [ "${arch}" == "${host_arch}" ]`) to keep the current behavior, else override with the proposed solution.

bdrewery (approved):
Seems fine. Do we really need 2 different methods thought?

jlduran:
> Seems fine. Do we really need 2 different methods thought?

Someone might want to have the hostname set on `${WRKDIR}/world/etc/rc.conf.d/hostname`, for example.

`sysrc -R ${WRKDIR}/world/` works perfectly, except when `${arch} != ${host_arch}`. I didn't want to introduce a regression.

Evidently, as you pointed out, having only one method would be ideal, so adding `bin/sh` and `bin/csh` to `HLINK_FILES` seemed like the way to go (currently done only for MIPS), but the comment:

>    Endian issues on mips/mips64 are not handling exec of 64bit shells
>    from emulated environments correctly.  This works just fine on ARM
>    because of the same issue, so allow it for now.

Threw me off a bit, and I haven't fully explored this option, nor I fully understand it yet (i.e. not sure if it does what I think).

In the meantime, I thought a sensible option would be to have the two methods.

I'll promote this draft to a pull request. Thank you!

PR #994, opened 2022-04-29
Make it clear that overlays must be poudriere ports trees

0mp:


igalic:
this leaves [`poudriere-image(8)`](https://man.freebsd.org/poudriere-image(8)) as an exception, where the overlay directory is just a general thing

PR #1000, opened 2022-05-21
Use recent FreeBSD versions in examples.

pauamma:
Not sure whether these require bumping .Dd, so I left it untouched.


igalic:
always bump `.Dd`

pauamma:
> always bump `.Dd`

Thanks. Going to wait for substantive review to limit noise.

PR #1002, opened 2022-06-18
image: Remove extra "/" in $OUTPUTDIR

jlduran:
No functional change intended.


jlduran:
Purely cosmetic change.
It avoids printing a double slash in the output image path.

PR #1013, opened 2022-09-09
New pkg do not needs devfs to be mounted when running in chroot

gearnode:
This was introduced in [2019](https://github.com/freebsd/poudriere/commit/5c0ca1b2960bc565176b9462a5f643664f10d1ad) when pkg open `/dev/null` after doing the chroot. In the [next release](https://github.com/freebsd/pkg/commit/5a0cc7f37e783a636e11f5499f5f8449ed318e0e) of pkg, this should not be required anymore as pkg will open `/dev/null` before doing the chroot.

Should not be merged until the new pkg is released.

### DIFF
--- a/src/share/poudriere/image.sh
+++ b/src/share/poudriere/image.sh
@@ -542,12 +542,10 @@ if [ -n "${PACKAGELIST}" ]; then
 		FreeBSD: { enabled: false }
 		local: { url: file:///tmp/packages }
 		EOF
-		mount -t devfs devfs ${WRKDIR}/world/dev
 		convert_package_list "${PACKAGELIST}" | \
 		    xargs chroot "${WRKDIR}/world" env \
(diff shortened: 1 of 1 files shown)


bapt:
the should not be merged until new pkg is release and available in the quarterly branch

bdrewery:
Poudriere doesn't require using the latest ports tree or `pkg`. For example, at work we can still use latest Poudriere with our older custom ports tree with older `pkg`. Is there harm in having this mount?

5u623l20:
> the should not be merged until new pkg is release and available in the quarterly branch
Not a poudriere question but I am curious to know whether if this is true as I was trying to build some custom images and pkg still needs devfs to be mounted. I was using HEAD(src) and a custom pkg repository built from HEAD(ports) and still needed devfs to be mounted. Even if the said change has not reflected into quarterly I believe it should be in the HEAD but despite that it's failing.

PR #1018, opened 2022-10-02
Remove zpool `altroot` when creating datasets.

kostaslambda:
`poudriere` breaks if the zpool it's given to use has the `altroot` property set, as its internal dataset mountpoints are different than the path they were actually mounted on.

Specifically, if the zpool has `altroot` set, all mounted datasets will have `altroot` prepended to their mountpoint. When setting `BASEFS`, the user would naturally take that into account and include `altroot` in the path. The setting would look something like: `BASEFS=<altroot>/rest/of/path`. However, what happens is that the `BASEFS` mountpoint is actually `<altroot><altroot>/rest/of/path`, while `poudriere` thinks it's `<altroot>/rest/of/path`.

To fix the problem, simply check if `altroot` is set and remove it from the mount path at the point of the dataset creation. The dataset will be prepended with `altroot`, which will match the path `poudriere` knows about.


kostaslambda:
ping?

src/share/poudriere/include/fs.sh
```diff
@@ -23,6 +23,15 @@
 # OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 # SUCH DAMAGE.
 
+remove_altroot() {
+	[ $# -ne 1 ] && eargs remove_altroot mountpoint
+	local mountpoint altroot
+	mountpoint="$1"
+	altroot="$([ -z "${NO_ZFS}" ] && zpool get -Ho value altroot "${ZPOOL}")"
```
bdrewery, line 30: Why the `$()` for this? We don't need a subshell for the test; we may avoid a fork if `NO_ZFS` is set.
We need to set a default value or unset `altroot` too for `NO_ZFS`.
```
altroot=
if [ -z "${NO_ZFS}" ]; then
  altroot=$(zpool get -Ho value altroot "${ZPOOL}")
fi
```
Or
```
altroot=
[ -n "${NO_ZFS}" ] || altroot=$(zpool get -Ho value altroot "${ZPOOL}")
```

The `test && test` will exit with `set -e`, even if in a subshell. Using `||` won't exit.
```
# sh -c 'set -e; foo=$(false && true); echo here'; echo $?
1
# sh -c 'set -e; foo=$(false && true) || :; echo here'; echo $?
here
0
# sh -c 'set -e; foo=$(false || true); echo here'; echo $?
here
0
```

bdrewery (changes requested):

bdrewery:
`altroot` seems like a temporary thing. What's the use case for something like this?

src/share/poudriere/include/fs.sh
kostaslambda: Done. Thanks for the detailed example. My shell-fu is not very strong :)

kostaslambda:
I came across this while trying to use poudriere on TrueNAS, which sets `altroot=/mnt` for all pools. Maybe it's uncommon for people to use that property, but it could still happen.

I've also changed the commit message, as it was wrong and maybe a bit convoluted. Let me know what you think.

PR #1027, opened 2022-11-21
Fix a typo in the zsh completion

JonyEpsilon:
I accidentally fired this miniature footgun earlier. Serves me right for being too hurried to double-check the manpage:-)


bdrewery:
Thanks!

PR #1029, opened 2022-12-11
Adding pushover support for notification

5u623l20:
- Rationalize the email part with more conditionals

Currently we are using pushover(https://pushover.net) to get notifications while on the road rather than email. This update just adds the relevant part for sending the pushover notifications. .


bdrewery:
Thanks!

PR #1036, opened 2023-01-15
image_zfs: destroy correct zpool on cleanup

rob-wing:
As of [image_zfs: Allow same-as-host pool names](https://github.com/freebsd/poudriere/commit/73b3a5169b640460c9bb8cbd295764e9e74838ab), if an error is encountered while building an image via `image_zfs.sh`, it's possible the wrong pool will be destroyed on cleanup. This happens because the on-disk name is used when destroying the pool instead of the in-core name.


src/share/poudriere/image_zfs.sh
```diff
@@ -38,10 +38,11 @@ _zfs_writereplicationstream()
 
 zfs_check()
 {
+	zroot="${ZFS_POOL_NAME}.$(jot -r 1 1000000000)"
```
jlduran, line 41: While here, what happens if someone overrides (e.g., via a pre-script) this `zfs_check()` function and "forgets" to set `zroot`?

jlduran (commented):
Nice, good catch. Thank you!

patmaddox:
I'll try running this over the next week. I had several failures today (on poudriere-devel from ports) where the message was "unable to destroy zroot" - which is a good thing it was unable, since that's my root pool! Then poudriere and any zfs commands would hang, and I had to reboot. So, I'm hopeful that this addresses that.

patmaddox:
I'm happy to report that using this branch, poudriere cleans up the correct zpool and doesn't try to destroy my zroot! Thank you :rocket:

allanjude:
ping

PR #1038, opened 2023-01-19
Use relative path for the EXCLUDELIST file too

ocochard:
Fixing the usage of relative file with -X option

### DIFF
--- a/src/share/poudriere/image.sh
+++ b/src/share/poudriere/image.sh
@@ -369,6 +369,10 @@ while getopts "A:bB:c:f:h:i:j:m:n:o:p:P:R:s:S:t:vw:X:z:" FLAG; do
 			SWAPSIZE="${OPTARG}"
 			;;
 		X)
+			# If this is a relative path, add in ${PWD} as
+			# a cd / was done.
+			[ "${OPTARG#/}" = "${OPTARG}" ] && \
+			OPTARG="${SAVED_PWD}/${OPTARG}"
 			[ -r "${OPTARG}" ] || err 1 "No such exclude list ${OPTARG}"
 			EXCLUDELIST=$(realpath ${OPTARG})
 			;;


src/share/poudriere/image.sh
```diff
@@ -369,6 +369,10 @@ while getopts "A:bB:c:f:h:i:j:m:n:o:p:P:R:s:S:t:vw:X:z:" FLAG; do
 			SWAPSIZE="${OPTARG}"
 			;;
 		X)
+			# If this is a relative path, add in ${PWD} as
+			# a cd / was done.
+			[ "${OPTARG#/}" = "${OPTARG}" ] && \
+			OPTARG="${SAVED_PWD}/${OPTARG}"
 			[ -r "${OPTARG}" ] || err 1 "No such exclude list ${OPTARG}"
 			EXCLUDELIST=$(realpath ${OPTARG})
```
bdrewery, line 377: Does this `realpath` invalidate the need for your fix?

ocochard: No, it need to have the same behaviour as all other poudriere-image option (like -P, -F, -B, etc.)

ocochard: Example with all configs files in local current directory named "poudriere.etc" (to avoid conflicting with system poudriere setup):

```
 # poudriere -e poudriere.etc image -t firmware -s 2g -j BSDRPj -p BSDRPp -n BSDRP -h router.bsdrp.net -c BSDRP/Files/ -f poudriere.etc/poudriere.d/BSDRP-pkglist -X poudriere.etc/poudriere.d/excluded.files -A poudriere.etc/poudriere.d/post-script.sh

[00:00:00] Error: No such exclude list poudriere.etc/poudriere.d/excluded.files

# cat poudriere.etc/poudriere.d/excluded.files
*.a
usr/sbin/config
usr/lib/debug
usr/sbin/bsdconfig
usr/share/bsdconfig
usr/share/bhyve
usr/share/examples
usr/share/flua
usr/share/doc
usr/share/man/man3
usr/share/man/man3lua
usr/share/man/man9
usr/share/mk
usr/share/openssl
usr/include
usr/share/misc/magic.mgc
usr/share/misc/termcap.db
usr/local/include
```

PR #1045, opened 2023-03-25
Allow '-f' argument to run in foreground

crees:
Using a config file option means that if it's set to foreground, the system will hang on startup if the RC script is enabled

If you'd prefer to use getopt(3) then I'm happy to rewrite like that.

### DIFF
--- a/src/poudriered/poudriered.c
+++ b/src/poudriered/poudriered.c
@@ -897,6 +897,10 @@ main(int argc, char **argv)
 
 	const ucl_object_t *sock_path_o, *pidfile_path_o, *foreground_o;
 
+	if (argc == 2 && strcmp(argv[1], "-f") == 0) {
+		foreground = true;
+	}
+
 	if (argc == 3) {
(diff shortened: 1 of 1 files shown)


src/poudriered/poudriered.c
```diff
@@ -897,6 +897,10 @@ main(int argc, char **argv)
 
 	const ucl_object_t *sock_path_o, *pidfile_path_o, *foreground_o;
 
+	if (argc == 2 && strcmp(argv[1], "-f") == 0) {
+		foreground = 1;
```
igalic, line 901: let's use `true` here. this *is* a boolean variable

igalic (changes requested):
eyes

src/poudriered/poudriered.c
crees: Well, that is sloppy isn't it, I'm sorry!  I'll fix it.

crees:
No idea why the first one failed- it's nothing to do with my patch!

PR #1046, opened 2023-03-25
Bump default `MAX_FILES` to 4096.

dag-erling:
This is necessary to build some large ports, e.g. `www/qt5-webengine` which since 5.15.8 fails at the linking stage with the current limit of 1024 (cf. [this patch](https://github.com/freebsd/freebsd-ports/blob/8067206f03a28c8eb926d07550d94220d7aacfbd/www/qt5-webengine/files/patch-src_core_core__module.pro)).


igalic (approved):
👍

mjguzik:
i would bump it to 8k for some future-proofing

the limit was put in place because by default freebsd allows literally millions of fds and at the time closefrom was not widely used, making hand-rolled "close on exec" loops incredibly nasty

from my tests iterations like that don't happen anymore apart from fakeroot and maybe some other program

PR #1051, opened 2023-05-06
Modernize some aspect of the HTML

5u623l20:
- Lots of HTML meta tags have ben removed since this HTML was created and are no logner available or supported in modern Browsers:
  - meta http-equiv="Cache-Control"
  - meta http-equiv="Pragma"
  - meta http-equiv="Expires"
  - It's advised to utilize HTTP header from the webserver to control the above
- Remove trailing slashes as trailing slash on void elements has no effect and interacts badly with unquoted attribute values.
- "_new" was never a standardized target value for HTML5. Replace that with "_blank"
- All script resources are now processed with charset "utf-8" and the attribute is not needed at all
- "type" attribute for JavaScript resources are unnecessary now
- Replace the local BootStrap, jQuery and DataTabels assets with the one available from cloud networks which reduces lag and increases accessibility on mobile networks


5u623l20:
Looks like this will need some more fixes. I will resubmit once I have fixed those.

PR #1063, opened 2023-07-23
allow native compliation for armv6 on armv7 arch as CPU-s are backward compatible

samm-git:
poudriere asks for QEMU binary if running on armv7 host and target is armv6. As ARMv7 is backward compatible with v6 it is not realy needed. This patch relaxes this check

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -2644,6 +2644,9 @@ need_emulation() {
 	# kern.supported_archs is a list of TARGET_ARCHs.
 	target_arch="${wanted_arch#*.}"
 
+        # armv6 binaries can natively execute on armv7, no emulation needed
+        [ "${target_arch}" = "armv6" ] && target_arch="armv[67]"
+
 	# Check the list of supported archs from the kernel.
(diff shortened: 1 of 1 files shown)


samm-git:
Ping :)

igalic:
@samm-git  can you fix the title of commit message / pull request?

samm-git:
@igalic done, i hope its better now

igalic:
still needs fixing in the commit message:

```
git commit --amend
git push -f $your-origin patch-1
```

samm-git:
@igalic done, thanks. Also rebased to the upstream

igalic (approved):
👍

bapt:
I don't like adding this "hack" to poudriere, kern.supported_arch is supposed to say armv6 if it does support it, imho we should fix the kernel to expose it. I need to think more about it.

samm-git:
@bapt it would be nice for the kernel to expose it, agree, but not sure that it's realistic, given the current state of the armv6 support. I am using this hack to build rpi1 packages and at least for the packages I am using it works reasonably well, much better compared to armv6/qemu on amd64 option IMO.

PR #1085, opened 2023-11-05
Update test-ports and remove svn

5u623l20:


bapt:
I will look into merging that one once 3.4.0 is out

Issue #1083, opened 2023-11-01
options command generating errors about missing ports tree FLAVOR feature


madpilot78:
## Prerequisites
- [x] Have you checked for an existing issue describing your problem?
- [x] Are you running the latest version?
- [x] Is your ports tree recent?
- [x] Is your FreeBSD Host on a [supported release](https://www.freebsd.org/security/#sup)?

## Describe the bug

With latest `ports-mgmt/poudriere-devel` port running the `poudriere options` command passinga  port with a flavor (f.e. `net/unison@nox11`) causes errors like:

```
[00:00:00] Error: Trying to build FLAVOR-specific net/unison@nox11 but ports tree has no FLAVORS support.
```

And the port is skipped.

## How to reproduce

see above

## Expected behavior
No errors being output. No skipping

## Environment
 - Host OS [e.g. 12.2 amd64]: head
 - Jail OS [e.g. 12.0 powerpc]: any
 - Poudriere Version [e.g. 3.3.1 or git hash or port version]: 3.4.99.20231030
 - Ports branch and revision [e.g. 2020Q3 r550754]: head

## Additional context

After further investigation, when running the options command, the P_PORTS_FEATURES variable is not being populated via the options command.

If I get more insight to share I'll do.

madpilot78:
The SUBPACKAGES feature is still being worked on, support needs to be added in poudriere before it can be unleashed on the ports tree.

But we should be getting it at some point.

Looks like the bug was introduced when adding support for SUBPACKAGES and not updating the options.sh script, but looking at it I found nothing obvious about how to fix this.

madpilot78:
> * Handle missing ports features gracefully.

This is in theory already done. Obviously if you ask for something that depends on a feature (like a specific flavor) there is not much to be done, skipping it is correct. Reverting on the unflavored port looks arbitrary.

Anyway the only issue at hand is the options subcommand not properly populating `$P_PORTS_FEATURES`.

I tried to figure it out but could not find a solution. I plan to find some spare time in the near future to try harder.

madpilot78:
I have created a pull request with a patch that works fine for me.

You can test it and report back.

Find it here: #1087

Thanks!

ahktenzero:
[...]

PR #1087, opened 2023-11-11
Try to fix `poudriere options` when specifying ports with flavor

madpilot78:
As reported in bug #1083 `poudriere options` is giving errors if it is passed ports specifying a flavor.

The `options.sh` script fails to populate the `P_PORTS_FEATURES` variable.

I have created this patch that starts a full jail so the `port_var_fetch` function can work and populate it.

This function is called via the `listed_ports` function call.

I'm not sure this is the correct approach, but is working fine here.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -45,7 +45,8 @@ was_a_testport_run() {
 # Return true if in a bulk or other jail run that needs to shutdown the jail
 was_a_jail_run() {
 	was_a_bulk_run ||  [ "${SCRIPTNAME}" = "pkgclean.sh" ] || \
-	    [ "${SCRIPTNAME}" = "foreachport.sh" ]
+	    [ "${SCRIPTNAME}" = "foreachport.sh" ] || \
+	    [ "${SCRIPTNAME}" = "options.sh" ]
 }
 schg_immutable_base() {
 	[ "${IMMUTABLE_BASE}" = "schg" ] || return 1
--- a/src/share/poudriere/options.sh
+++ b/src/share/poudriere/options.sh
@@ -193,6 +193,11 @@ options_cleanup() {
 }
 setup_makeconf ${__MAKE_CONF} "${JAILNAME}" "${PTNAME}" "${SETNAME}"
 
+# to cycle ports this command calls the listed_ports function which
+# needs to check for features.
+# This check needs the P_PORTS_FEATURES to be populated by
+# fetch_global_port_vars which requires a running jail to work properly.
+# So provide a running jail here.
 MASTERNAME=${JAILNAME}-${PTNAME}${SETNAME:+-${SETNAME}}
 _mastermnt MASTERMNT
 
--- a/src/share/poudriere/options.sh
+++ b/src/share/poudriere/options.sh
@@ -190,7 +190,6 @@ export __MAKE_CONF
 CLEANUP_HOOK=options_cleanup
 options_cleanup() {
 	rm -f ${__MAKE_CONF}
-	jail_stop
 }
 setup_makeconf ${__MAKE_CONF} "${JAILNAME}" "${PTNAME}" "${SETNAME}"
 
--- a/src/share/poudriere/options.sh
+++ b/src/share/poudriere/options.sh
@@ -190,6 +190,7 @@ export __MAKE_CONF
 CLEANUP_HOOK=options_cleanup
 options_cleanup() {
 	rm -f ${__MAKE_CONF}
+	jail_stop
 }
 setup_makeconf ${__MAKE_CONF} "${JAILNAME}" "${PTNAME}" "${SETNAME}"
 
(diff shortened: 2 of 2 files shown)


madpilot78:
> It does, however, have trouble if you happen to `^c` during options dialog config. Doesn't clean up the jails...

Oh, I see, did not think about this. I will need a little time to figure this one out.

madpilot78:
I added a call to `jail_stop` to the `CLEANUP_HOOK` and this does fix the `^C` issue, although, again, I'm not sure this is the correct way to achieve it. I see no other script doing this.

I'm open to suggestions for improvements though.

madpilot78:
@vanvoljg I agree adding some comments would be useful, I'll add a commit soon with short comments explaining my changes.

BTW how a shell script can run code to cleanup when interrupted itself is easily explained: the use the trap shell builtin. But, being poudriere a complex script it has its own architecture of trap functions also extensible via the `CLEANUP_HOOK` variable, which I am not sure I understand the inner workings of.

madpilot78:
I'm starting to make sense of it. In theory the `exit_handler`, bound to run on exit via `trap`, calls `jail_cleanup` [here](https://github.com/madpilot78/poudriere/blob/e238ce427236a94b34c0e5281fb2525ba635a4da/src/share/poudriere/common.sh#L1359). But before doing that, it checks the `STATUS` variable.

Such variable should be set in `jail_start` [here](https://github.com/madpilot78/poudriere/blob/e238ce427236a94b34c0e5281fb2525ba635a4da/src/share/poudriere/common.sh#L3053), depending on the `SET_STATUS_ON_START` variable (which I don't think I'm setting, so the default of `1` should be used).

It looks like this is not happening, which would explain why the jail is not correctly being cleaned up.

SO correct solution would be to make this already existing mechanism work as expected, and not adding things to `options_cleanup`. I need to understand why STATUS is not being set properly (if that is really the issue)

madpilot78:
@vanvoljg 

My analysis in comment https://github.com/freebsd/poudriere/issues/1087#issuecomment-1807275749 was incomplete. STATUS is correctly set. But, in `jail_cleanup`, at:

https://github.com/freebsd/poudriere/blob/234f8d15a4ca61b71e2a55364cca695906fe58b0/src/share/poudriere/common.sh#L3488

The code also checks this `was_a_jail_run` function output (find it at the start of common.sh). So I fixed that and everything now works as expected.

Being poudriere a shell script i don't think it is possible to clear really all cases. A `^C` hit at the wrong time could end up interrupting the script in the wrong moment while things are being setup (non atomically) for automatic cleanup.

Major refactoring of this code would be a very complex task that could cause more issues than it resolves, and I'm not an expert of poudriere code structure, so I would not feel confident taking on such a task. I am more in favour of an incremental strategy to improving it, at least for now.

Also, if `^C` is hit two times in a row, the celanup code will be interrupted.

bdrewery:
I haven't looked at this in detail but had this commit pending to push. I think having a jail running now is better anyway. Note I did add `fetch_global_ports_vars` in a few places that still may be needed.
```
commit aac72a85ca538bd790b7294424b219c3e5490b8e
Author: Bryan Drewery <bryan@shatow.net>
Date:   Wed Nov 30 20:13:06 2022 -0800

    Fix some fallout from removing DEPENDS_ARGS.

    The have_ports_feature() was broken before and did not properly
    have FLAVORS set where listed_ports() was used.

diff --git src/share/poudriere/distclean.sh src/share/poudriere/distclean.sh
index f8ae54fb1..841df6879 100755
--- src/share/poudriere/distclean.sh
+++ src/share/poudriere/distclean.sh
@@ -154,6 +154,8 @@ for PTNAME in ${PTNAMES}; do
                export PACKAGE_BUILDING=yes
                echo "PACKAGE_BUILDING_FLAVORS=yes"
        fi >> "${__MAKE_CONF}"
+       unset P_PORTS_FEATURES
+       fetch_global_port_vars

        MASTERMNT= MASTERMNTREL= load_moved
        msg "Gathering all expected distfiles for ports tree '${PTNAME}'"
diff --git src/share/poudriere/foreachport.sh src/share/poudriere/foreachport.sh
index 993f3fec6..e1e831d9e 100755
--- src/share/poudriere/foreachport.sh
+++ src/share/poudriere/foreachport.sh
@@ -149,6 +149,7 @@ run_hook foreachport start
 exec >&3

 export PORTSDIR
+fetch_global_port_vars
 clear_dep_fatal_error
 parallel_start
 for originspec in $(listed_ports show_moved); do
diff --git src/share/poudriere/options.sh src/share/poudriere/options.sh
index 745c20d80..9c933937d 100755
--- src/share/poudriere/options.sh
+++ src/share/poudriere/options.sh
@@ -51,6 +51,9 @@ Options:
 EOF
        exit ${EX_USAGE}
 }
+injail() {
+       "$@"
+}

 ARCH=
 PTNAME=default
@@ -186,6 +189,7 @@ options_cleanup() {
        rm -f ${__MAKE_CONF}
 }
 setup_makeconf ${__MAKE_CONF} "${JAILNAME}" "${PTNAME}" "${SETNAME}"
+fetch_global_port_vars

 export TERM=${SAVED_TERM}
 for originspec in $(listed_ports show_moved); do
```

PR #1093, opened 2023-11-27
fix blacklist loading

lherschi:


lherschi:
Since no one here wanted to commit it, the error has now crept into the new release.

PR #1094, opened 2023-11-27
Add new poudriere audit command to check repositories for vulnerable …

brd:
…packages

### DIFF
--- a/src/share/poudriere/audit.sh
+++ b/src/share/poudriere/audit.sh
@@ -1,0 +1,83 @@
+#!/bin/sh
+#
+# Copyright (c) 2023 Brad Davis <brd@FreeBSD.org>
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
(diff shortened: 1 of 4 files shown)


brd:
@bdrewery Do you have any thoughts on this?

src/share/poudriere/audit.sh
```diff
@@ -0,0 +1,83 @@
...
+
+MASTERNAME=${JAILNAME}-${PTNAME}${SETNAME:+-${SETNAME}}
+PACKAGES="${POUDRIERE_DATA:?}/packages/${MASTERNAME}"
+_mastermnt MASTERMNT
+
+PKG_EXT='*' package_dir_exists_and_has_packages || \
+	err 0 "No packages exist for ${MASTERNAME}"
+
+msg "Auditing for jail '${JAILNAME}'"
+if ! ${PKG_BIN} audit -d "${PACKAGES}"; then
```
bdrewery, line 81: What sets `PKG_BIN`? I think we should just use `pkg` here.

bdrewery (changes requested):

bdrewery:
I like it!

brd:
I forgot about the pkg side of the changes.. see freebsd/pkg#2247

src/share/poudriere/audit.sh
brd: Done.

bdrewery:
Hmm I still see `${PKG_BIN}` in latest commit.

brd:
Yup, forgot to commit it.. Done now

bdrewery (approved):

bdrewery:
Let's get the `pkg` change in first before merging this?

bapt:
the proposed patch for pkg would have never worked however in git pkg now has support for audit -d

https://github.com/freebsd/pkg/commit/5860bc22ede55eebe37769205ae8ca7014c89a24

allanjude:
Is this ready to land now? or do we need to refresh it?

PR #1101, opened 2023-12-25
Fix broken deps sometimes, when using of -S

dsh2dsh:
Hi,

I'm not sure how correct my solution is, but may be it'll be better.

Without `-S` `bulk` removes the full chain of dependencies. For instance:

```
Deleting vulkan-loader-1.3.273.pkg: new version: 1.3.274
Deleting ffmpeg-6.1_1,1.pkg: missing dependency: vulkan-loader-1.3.273
Deleting firefox-121.0,2.pkg: missing dependency: ffmpeg-6.1_1,1
```

`firefox-121.0,2` doesn't depend on `vulkan-loader-1.3.273`, but `bulk` removes
it too, because `ffmpeg-6.1_1,1` was removed.

With `-S` `bulk` just skips this step at all and we have `ffmpeg-6.1_1,1`, which
now depends on nonexisted `vulkan-loader-1.3.273`, because
`vulkan-loader-1.3.273` was removed.

Now, with changed `-S`, `bulk` will remove `ffmpeg-6.1_1,1`, but keep
`firefox-121.0,2`, because version of `ffmpeg` will not change after rebuilding
and `firefox-121.0,2` will have valid dependencies. The same step now looks:

```
Deleting vulkan-loader-1.3.273.pkg: new version: 1.3.274
(-S) Keeping ffmpeg-6.1_1,1.pkg: missing dependency: vulkan-loader-1.3.273

[ ... after processing dependencies ]

(-S) Deleting ffmpeg-6.1_1,1: missing dependency
```

Also I moved `download_from_repo` below `sanity_checks_pkgs`, because `sanity_check_pkgs` removes packages which don't have builded dependencies and if we'll `download_from_repo` before `sanity_check_pkgs`, we will rebuild it. But if we'll `download_from_repo` after `sanity_check_pkgs`, we could download it instead of rebuild, which is faster.


bdrewery:
Thanks for this. Unfortunately the plan is to remove `-S` and make its goal the default, with a _lot_ more logic than here to handle the needed rebuild cases. #822 handles that. `-S` fundamentally never worked and was a mistake.

PR #1102, opened 2023-12-27
Add -y to the ports subcommand to remove the confirmation prompt

brd:
…hout confirmation


bapt:
please update the manpage as well

brd:
Good catch.  I have updated the man page

igalic (approved):

PR #1103, opened 2023-12-29
timestamp.c: handle error if memory allocation fails and a few others

rilysh:
Please see the [commit message](https://github.com/freebsd/poudriere/commit/24f74e83226f46d02e08a94d5152852466604ddf).


igalic (approved):
this seems sensible 👍

PR #1108, opened 2024-01-05
Ignore automatically generated files by autoconf and automake

fel1x-developer:
With change, developers can use newer versions of Autotools when it is not updated on FreeBSD ports. 
Internal changes in `Makefile.in`, `m4/`, etc will be ignored when `./autogen.sh` and `./configure`


bapt:
I actually plan to remove entirely autotools, (not fully decided yet) so I will take time to think about it more before merging this

bdrewery:
build-aux/test-driver shouldn't be removed as it has customizations for the test framework.
I don't like the route here. Poudriere is supposed to be self-sufficient but this creates a dependency for no benefit.

fel1x-developer:
> I don't like the route here. Poudriere is supposed to be self-sufficient but this creates a dependency for no benefit. As for switching away from autotools I don't see the point there either. What we have works and the test framework is integrated into it.

As autoconf and automake are already required, I don't think many users will be bothered by this change. libtool is a part of autotools (and also devel/autotools metaport), and autoconf-archive is also accessible to FreeBSD users with pkg or ports.

The benefit of this way is that we no longer need to update test-drivers/ and m4/ from the upstream autoconf-archive manually. Also, I don't think it's a good practice to keep automaticallly generated file under version control.

[StackOverflow thread about this topic](https://stackoverflow.com/questions/3290908/which-files-generated-by-autotools-should-i-keep-in-version-control-repository)

fel1x-developer:
Since `test-driver` is customized for our own use, I agree that it should be tracked under version control.

fel1x-developer:
Fix merge conflict

.gitignore
```diff
@@ -1,3 +1,61 @@
...
+/autoscan-*.log
+/aclocal.m4
+/compile
+/config.cache
+/config.guess
+/config.h.in
+/config.log
+/config.status
+/config.sub
+/configure
```
igalic, line 26: I think i would keep configure in git for convenience

igalic (changes requested):

.gitignore
fel1x-developer: `configure` file is machine-dependent (e.g. latest autoconf version can differ in different platforms), and it has to be updated manually on git. If someone forgets to update ./configure with autogen.sh and only push changes on autogen.sh, this might cause inconvinience for other devs, too.

If there is a way to always keep `configure` up-to-date, I will leave it out from .gitignore

bdrewery:
> As autoconf and automake are already required,

They are not. That's the point of having all of these files checked in. The change would require dependencies for Poudriere, for which they are none right now.
Thank you for the contribution but we're going to keep this as is right now. If there's a problem that needs to be addressed let's focus on that.

PR #1111, opened 2024-01-20
Write port version and port revision in logs

yurivict:
Port version will be useful for collecting analytics information from poudriere logs.


igalic (approved):
👍🏻

PR #1116, opened 2024-01-30
jail: Fix pkgbase jail creation

evadot:
Since the official repo aren't using 'latest' directory fix how we construct the pkgbase url.
Now it's flexible enough to accomodate for most use case. An example on how to create a 15-CURRENT jail:
poudriere jail -j pkgbase-main-weekly -c -m pkgbase=base_weekly -U https://pkg.freebsd.org/ -v 15 -a amd64

This will use the base_weekly repository, while
poudriere jail -j pkgbase-main-weekly -c -m pkgbase=base_latest -U https://pkg.freebsd.org/ -v 15 -a amd64 will use the base_latest one.

This also install the FreeBSD-src-sys package so we can build kmods in the jail.

Sponsored by:	Beckhoff Automation GmbH & Co. KG

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -8438,6 +8438,17 @@ svn_git_checkout_method() {
 				;;
 			esac
 			;;
+		pkgbase*)
+			case "${SOURCES_URL}" in
+			http://*) ;;
+			https://*) ;;
+			file://*) ;;
+			*)
+				msg_error "Invalid pkgbase url"
+				return 1
+				;;
+			esac
+			;;
 		*)
 			msg_error "-U only valid with git and svn methods"
 			return 1
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -63,14 +63,14 @@ Options:
                      obtaining and building the jail. See poudriere(8) for more
                      details. Can be one of:
                        'ftp-archive', 'ftp', 'freebsdci', 'http',
-		       'null', 'src=PATH', 'tar=PATH', 'url=URL', or
+		       'null', 'src=PATH', 'tar=PATH', 'url=URL', 'pkgbase=repo' or
 		       '{git,svn}{,+http,+https,+file,+ssh}' (e.g., 'git+https').
                      The default is '${METHOD_DEF}'.
     -P patch      -- Specify a patch to apply to the source before building.
     -S srcpath    -- Specify a path to the source tree to be used.
     -D            -- Do a full git clone without --depth (default: --depth=1)
     -t version    -- Version of FreeBSD to upgrade the jail to.
-    -U url        -- Specify a url to fetch the sources (with method git and/or svn).
+    -U url        -- Specify a url to fetch the sources (with method git, svn and pkgbase).
     -X            -- Do not build and setup native-xtools cross compile tools in jail
                      when building for a different TARGET ARCH than the host.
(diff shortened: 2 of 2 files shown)


src/share/poudriere/common.sh
```diff
@@ -8449,6 +8449,17 @@ svn_git_checkout_method() {
 				;;
 			esac
 			;;
+		pkgbase*)
+			case "${SOURCES_URL}" in
+			http://*) ;;
```
igalic, line 8454: does `http+pkg://` not work?

src/share/poudriere/jail.sh
```diff
@@ -857,11 +857,12 @@ FreeBSD: {
 }
 EOF
 
-	pkg -o REPOS_DIR="${JAILMNT}/etc/pkg" -r ${JAILMNT}/ update
-	# Omit the man/debug/kernel and tests packages, uneeded for us.
-	pkg search -qx '^FreeBSD-.*' | grep -vE -- '-man|-dbg|-kernel-|-tests' | xargs pkg -o REPOS_DIR="${JAILMNT}/etc/pkg" -r ${JAILMNT}/ install -y
+	pkg -o REPOS_DIR="${JAILMNT}/etc/pkg" -o ABI="FreeBSD:${VERSION}:${ARCH}" -r ${JAILMNT}/ update
+	# Omit the man/debug/kernel/src and tests packages, uneeded for us.
+	pkg -o REPOS_DIR="${JAILMNT}/etc/pkg" -o ABI="FreeBSD:${VERSION}:${ARCH}" -r ${JAILMNT}/ search -qCx '^FreeBSD-.*' | grep -vE -- '-man|-dbg|-kernel-|-tests|-src-' | xargs pkg -o REPOS_DIR="${JAILMNT}/etc/pkg" -r ${JAILMNT}/ install -y
```
igalic, line 862: I would add `-w` to `grep`, just to be on the safe side

src/share/poudriere/jail.sh
```diff
@@ -63,14 +63,14 @@ Options:
...
-		       'null', 'src=PATH', 'tar=PATH', 'url=URL', or
+		       'null', 'src=PATH', 'tar=PATH', 'url=URL', 'pkgbase=repo' or
 		       '{git,svn}{,+http,+https,+file,+ssh}' (e.g., 'git+https').
                      The default is '${METHOD_DEF}'.
     -P patch      -- Specify a patch to apply to the source before building.
     -S srcpath    -- Specify a path to the source tree to be used.
     -D            -- Do a full git clone without --depth (default: --depth=1)
     -t version    -- Version of FreeBSD to upgrade the jail to.
-    -U url        -- Specify a url to fetch the sources (with method git and/or svn).
+    -U url        -- Specify a url to fetch the sources (with method git, svn and pkgbase).
```
igalic, line 73: I realise this is unrelated to this Pull Request, but does it still make sense to keep supporting?

src/share/poudriere/jail.sh
evadot: Pretty sure that -w doesn't change a thing on this case.

src/share/poudriere/common.sh
evadot: probably does, let's first fix this feature and we can later expand it.
Right now it's not possible to give a full pkg repo description (with keys etc ...) so we might change again the format later.

PR #1117, opened 2024-01-31
Remove jail -c -m allbsd

brooksdavis:
It has been over 9 years since allbsd.org was updated.  The last snapshots are for HEAD post 10.0 and not particularly useful.  If it starts updating again someday this code can be restored.  For now it just gives false hope.


igalic (commented):
In a similar vein: svn has stopped updating, and most likely never still update again

PR #1122, opened 2024-02-10
fix git url documentation in poudriere.conf.sample

svmhdvn:


src/etc/poudriere.conf.sample
```diff
@@ -89,13 +89,13 @@ USE_TMPFS=yes
 DISTFILES_CACHE=/usr/ports/distfiles
 
 # If set the ports tree marked to use git will use the defined
```
bdrewery, line 91: Missed a "ports" here.

bdrewery (changes requested):

src/etc/poudriere.conf.sample
svmhdvn: Thanks!

bdrewery (approved):

Issue #1123, opened 2024-02-13
Missing file /var/run/os-release (generated by /etc/rc.d/os-release) in builder jail


ocochard:
## Describe the bug
<A clear and concise description of what the bug is.>
A new port is failing to build because is need to read  /var/run/os-release is missing (but symlink /etc/os-release is existing).

## How to reproduce
Run a testport in interactive mode and try to read /etc/os-release:

```
root@builder-official:# cat /etc/os-release
cat: /etc/os-release: No such file or directory
root@builder-official:# ls -alh /etc/os-release
lrwxr-xr-x  1 root wheel   21B Feb 12 23:29 /etc/os-release -> ../var/run/os-release
root@builder-official:# ls -alh /var/run/os-release
ls: /var/run/os-release: No such file or directory
```

## Expected behavior
rc script os-release should be started in the builder jail
Running `service os-release start` in the interactive jail fix it.

## Environment
 - Host OS : amd64
 - Jail OS: amd64
 - Browser: [e.g. chrome, safari]: 
 - Poudriere Version: poudriere-devel-3.4.99.20240122 
 - Ports branch and revision: main

PR #1124, opened 2024-02-13
Update common.sh

ocochard:
Some build system try to extract OS values from /etc/os-release, so populate it.

This is a workaround for [issue 1123](https://github.com/freebsd/poudriere/issues/1123).
But I don’t understand why it is not started by default (as instructed in /etc/defaults/rc.conf).


bdrewery:
The reason is we don't run `/etc/rc` during jail start.

PR #1130, opened 2024-02-29
Make mandir option of configure script work as is expected

yasuhirokimura:
Currently mandir is always $(prefix)/man even if other directory is specified with mandir option of configure script. So make the option work as is expected.


bdrewery:
8e48507e122f0a8f5403d4b138c6e54bf77243d7 obsoletes this. I'll remove the mandir line entirely too. Thanks for the PR.

PR #1143, opened 2024-04-04
Set a GPT label for EFI partition

ocochard:
This allow to easly update the EFI loader and sync with a standard FreeBSD install.

### DIFF
--- a/src/share/poudriere/image_firmware.sh
+++ b/src/share/poudriere/image_firmware.sh
@@ -134,7 +134,7 @@ firmware_generate()
 	espfilename=$(mktemp /tmp/efiboot.XXXXXX)
 	make_esp_file ${espfilename} ${ESP_SIZE} ${WRKDIR}/world/boot/loader.efi
 	mkimg -s gpt -C ${IMAGESIZE} -b ${mnt}/boot/pmbr \
-		-p efi/efiboot:=${espfilename} \
+		-p efi/efiboot0:=${espfilename} \
 		-p freebsd-boot:=${mnt}/boot/gptboot \
 		-p freebsd-ufs/${IMAGENAME}1:=${WRKDIR}/raw.img \
 		-p freebsd-ufs/${IMAGENAME}2:=${WRKDIR}/raw.img \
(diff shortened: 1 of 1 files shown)


src/share/poudriere/image_firmware.sh
```diff
@@ -134,7 +134,7 @@ firmware_generate()
 	espfilename=$(mktemp /tmp/efiboot.XXXXXX)
 	make_esp_file ${espfilename} ${ESP_SIZE} ${WRKDIR}/world/boot/loader.efi
 	mkimg -s gpt -C ${IMAGESIZE} -b ${mnt}/boot/pmbr \
-		-p efi:=${espfilename} \
+		-p efi/efiboot:=${espfilename} \
```
jlduran, line 137: I have been using `efiboot0` to keep it just like the output of bsdinstall (zfsboot), confirmed with: `gpart show -l`.
Although, the label appears as `efiesp` for [vimages](https://github.com/freebsd/freebsd-src/blob/d3b6d70e0dfaa4e9496bddcbc212d7d3ea392800/release/tools/vmimage.subr#L304), there is no standard yet, so this comment can be safely ignored.

jlduran: EDIT: I have submitted a small rant on the topic: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=278480 .

bsdimp: I think we should standardize with what the installer is doing: `efiboot0` It's likely the most widely deployed...

jlduran (approved):

evadot:
We probably should use the same name as the installer and iirc we don't use "efiboot0" there.

bsdimp:
I believe we're standardizing on efiboot0 everywhere.
IIRC old versions of bsdinstall used efiboot, but when multi-esp support was added, it was added as efiboot%d starting at 0.

evadot:
> I believe we're standardizing on efiboot0 everywhere. IIRC old versions of bsdinstall used efiboot, but when multi-esp support was added, it was added as efiboot%d starting at 0.

Yes indeed.

PR #1147, opened 2024-04-19
Makefile.am: Move optimization level to AM_CFLAGS

fel1x-developer:
Defining CFLAGS in Makefile.am is not recommended and generates a warning in autoreconf stage. Since CFLAGS is a user variable, it should not be directly modified by autoreconf.

Warning generated by `./autogen.sh`
```
Makefile.am:12: warning: 'CFLAGS' is a user variable, you should not override it;
Makefile.am:12: use 'AM_CFLAGS' instead
```

This fixes commit 52fff5ef4bcf8fa67dc90819494b9a43b9131122

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,8 +8,7 @@ if MAINTAINER_MODE
 AM_CFLAGS+=	-Wextra -Werror
 endif
 if ADDRESS_SANITIZER
-AM_CFLAGS+=	-fsanitize=address -DADDRESS_SANITIZER -fno-omit-frame-pointer
-CFLAGS+=	-O1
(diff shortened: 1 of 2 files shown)


igalic (approved):
👍

fel1x-developer:
Can this PR be comitted to the main branch?

allanjude (approved):

bdrewery:
Did you analyze how this changes the build? I see a change that is not documented here. That makes the -O1 do nothing. I wonder why this was purposely in `CLFAGS` and not the obvious line right next to it named `AM_CFLAGS`?

bdrewery:
This causes -01 to bot be used; -00 is used instead. -01 was added intentionally.

fel1x-developer:
Figured out that optimization level cannot be set through AM_CFLAGS. I'm withdrawing this PR.

PR #1148, opened 2024-04-21
Fetch packages after removing what we need to rebuild, not before that.

dsh2dsh:
I think, this change from PR #1101 is still actual:

https://github.com/freebsd/poudriere/pull/1101
https://github.com/freebsd/poudriere/pull/1101/commits/bc6b07bd3beebd89c238d677e99843d1519efc4a


bdrewery:
Yes the plan is for #822 to address. Or at least come in first before making more changes to the ordering.

PR #1158, opened 2024-06-18
poudriere.css: Adapt to latest DataTables style to avoid unreadable tables

### LINKED ISSUE #1114: Webinterface tables are hard to read due to background color
## Prerequisites
- [x] Have you checked for an existing issue describing your problem?
- [x] Are you running the latest version?
- [x] Is your ports tree recent?
- [x] Is your FreeBSD Host on a [supported release](https://www.freebsd.org/security/#sup)?

## Describe the bug
The current color scheme (`even` rows) for tables makes it really hard to read `build`/`failed`/`skipped`/`fetched` and `queued`.

## How to reproduce
1. start a build
2. open the webinterface
3. go to a build

## Expected behavior
Don't set the css `background-color` for `even` rows to `transparent`

## Screenshots
build
failed
skipped
fetched
queued

## Environment
 - Host OS: `13.2 amd64`
 - Jail OS: `13.2 amd64`
 - Browser: `Firefox` & `Chrome`
 - Poudriere Version: `3.4.1`
 - Ports branch and revision: `latest`

jrtc27:
The newer DataTables imported in ce2146e99efc ("Update DataTables to
1.13.8") started setting every other row to transparent. Since we add
the built etc classes to the whole table, this means the entire table's
background colour is green etc, and so the transparent rows now allow
that colour to shine through, making it hard to read the text.

Adapt the CSS selectors to only colour the thead within the whole table
so that the rows are restored to being uncoloured, less garish and
rather more readable. Note that these classes are also used for the
status display at the top, so we need to handle them on things other
than tables too, where we continue to just colour the whole element.

Fixes #1114

Fixes:	ce2146e99efc ("Update DataTables to 1.13.8")

### DIFF
--- a/src/share/poudriere/html/assets/poudriere.css
+++ b/src/share/poudriere/html/assets/poudriere.css
@@ -48,43 +48,43 @@
 .layout td {
     border: 0;
 }
-.built {
+.built:not(table), table.built thead {
     background-color: #00cc00 !important;
 }
 .built .row1 {
     background-color: #00ff00;
 }
-.failed {
+.failed:not(table), table.failed thead {
     background-color: #e00000 !important;
 }
 .failed .row1 {
     background-color: #ff0000;
 }
-.skipped {
+.skipped:not(table), table.skipped thead {
     background-color: #cc6633 !important;
 }
 .skipped .row1 {
     background-color: #aa6734;
 }
-.ignored {
+.ignored:not(table), table.ignored thead {
     background-color: #ff9900 !important;
 }
 .ignored .row1 {
     background-color: #ff6f00;
 }
-.fetched {
+.fetched:not(table), table.fetched thead {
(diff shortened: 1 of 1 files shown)


bdrewery:
@fel1x-developer thoughts?

fel1x-developer:
@jrtc27 Can you post screenshots of before/after?

Currently I cannot access to my FreeBSD instance right now, and I will be able to access to it by the end of August. However, if I see that the changes in the CSS file is clearly reflected in screenshots, I will accept this PR.

jrtc27:
I’m away from my laptop until the end of this week, but can provide screenshots when back if still needed

fel1x-developer:
I'm good with these changes

jrtc27:
Ping

jrtc27:
Ping

PR #1161, opened 2024-07-08
Update image_firmware.sh

ocochard:
Fix missing unterminated string


ocochard:
will create a new one

Issue #1129, opened 2024-02-26
Option to fail `poudriere bulk -b` if ports listed in `PACKAGE_FETCH_WHITELIST` are skipped


svmhdvn:
## Prerequisites
- [x] Have you checked for an existing issue describing your idea?

## What is your proposal?
Have a `poudriere.conf` option such as `PACKAGE_FETCH_WHITELIST_FAIL_ON_SKIP=yes` to cause any packages listed in `PACKAGE_FETCH_WHITELIST` to fail the entire build if they are skipped while fetching.

## What is the existing behavior, if any?
The build does mention the reason why fetching is skipped, but it silently continues building the unfetched ports.

## What is the motivation / use case for the change?
I have a poudriere server that is as powerful as I can afford right now. It still takes multiple hours to build massive ports such as `lang/rust` due to CPU and RAM constraints. `lang/rust` seems to be more and more prevalent as a build dependency for many ports I use in my pkglist.

Instead of trying to build this from source and wait hours before successful `poudriere bulk` builds, I'd like to fetch the binary instead, since `lang/rust` is just a build-time dependency for my desired ports only. However, due to Rust's relatively fast upstream release cadence and FreeBSD's official pkg repo `latest` branch cadence, it takes a while before the newest version of `lang/rust` hits the official pkg repo. This causes my `poudriere bulk -b` build to skip fetching `lang/rust` due to a local/remote ports version mismatch (since I keep my local checkout of the ports repo up-to-date).

`lang/rust` is the only port in my `PACKAGE_FETCH_WHITELIST`. I'd like to have the option to fail my `poudriere bulk -b` run when this package gets skipped while fetching. This would allow me to manually go to my local ports checkout, switch to a previous version of `lang/rust` that matches the latest official pkg repo version, and rebuild. This will save me hours of time and quite a bit of power usage as well from avoiding the heavy build of `lang/rust`.

## Did you consider any alternatives?
* FreshPorts is currently working on a feature to help its users watch for new package builds to be released to their respective ABI location in the official pkg repos. This is being tracked in https://github.com/FreshPorts/freshports/issues/542 . Once this feature is available, it should be easy enough to be notified when my particular case will be resolved. However, it's still better to just have the option to fail the build and know the reason for it.

## Is this really a ports feature request?
N/A

## Example
N/A

## Additional context
N/A

igalic:
honestly, even with freshports bringing the promised feature, i would still like to fail my build before building any compiler suite

[...]

PR #1164, opened 2024-07-27
First draft for a Fetch Only check

hauptsignal:
This is a first draft for a Fetch Only check funtion. It should offer the functionality described in Issue #1129 .

I have developed the function here on my locally installed version (3.4.1_1) of poudriere. I did a few small tests (undefined configuration, empty list, filled list, entries with asterisk), which at least worked for me.

But I have too little experience in developing poudriere. So I don't know if the function call is in a suitable place. And whether this installation does not interfere with other functions. Therefore I ask you to take a very close look at the whole thing and test it before integration.


hauptsignal:
Okay. My attempt does not work, if all the packages have to be rebuilt after a jail upgrade.

I think the Fetch Only check has to be at a later stage in the process. Maybe just before or at the start of the builder, that would build a mentioned port?

Issue #59, opened 2014-09-18
Security: Jails should have devfs rules applied to them


bapt:
After creating the jail, and mounting devfs, the devfs jail ruleset should be applied to the jail.  Otherwise, package builds essentially have full access to the host, and are easily able to escape the jail.

For example:

ls -l /poudriere/jails/83i386/build/09/dev
[...]
crw-r-----  1 root  operator  0x6b Jan 16 13:29 da0
[...]
crw-r-----  1 root  kmem      0x10 Jan 16 13:29 mem

So, the build jails have access to attached disks, and the kernel memory space.  Either can provide an escalation of privilege.

Pre-canned devfs rulesets are available in /etc/defaults/devfs.rules.  These hide all devices except for those that are safe inside jails, and should probably be used rather than rolling your own.  You may also be able to reuse devfs_mount_jail() etc from /etc/rc.subr.  Looking at how /etc/rc.d/jail does things may be useful.

Gavin

PR #1175, opened 2024-09-28
Allow using rulesets when mounting devfs for a jail

arrowd:
Specifically, this makes `/dev/zfs` available in the jail so that `zfs list` do not return error.


jlduran:
Per the original comment in #59:

> Pre-canned devfs rulesets are available in /etc/defaults/devfs.rules. These hide all devices except for those that are safe inside jails, and should probably be used rather than rolling your own.

I guess, this change is OK, as initially intended.

PR #1176, opened 2024-10-03
Add support for non-rsa PUBKEY signatures

kevans91:
Recent pkg accepts keys in the format "/path" or "${type}:/path", where the latter incurs a lookup in the pkgsign table for a supported signer to use the key.  Unprefixed paths are assumed to be RSA to maintain a semblance of backwards compatibility.

Add a pair of functions to parse out the components of the PKG_REPO_SIGNING_KEY so that one can just use the same key spec that pkg would accept.  We don't try to do any validation here so that we can just leave the source of truth of what's accepted in pkg to offer some flexibility in the future.

This is carefully designed to avoid using pkg-key(8) or a keyfile prefix if we're just using rsa, so that we don't force a new pkg version on the user until they want ecc support.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -205,6 +205,38 @@ yes) ;;
 	alias dev_err='# ' ;;
 esac
 
+# Extract key type from PKG_REPO_SIGNING_KEY.
+key_type() {
+	local keyspec="${PKG_REPO_SIGNING_KEY}"
+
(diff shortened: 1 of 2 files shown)


kevans91:
Closing for now; I'll finish up pkg(7) support first, then reopen

kevans91:
pkg(7) support stack starting at https://reviews.freebsd.org/D48104

PR #1180, opened 2024-10-15
jail: Use the right arch for a pkgbase jail

jlduran:
This is currently a problem for anything not `amd64` when building pkgbase jails.  Strip the `ARCH` and use only what corresponds to `TARGET_ARCH`.

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -372,9 +372,9 @@ update_jail() {
 	pkgbase)
 		VERSION=$(jget ${JAILNAME} version | cut -d '.' -f 1)
 		[ -z "${ARCH}" ] && ARCH=$(jget ${JAILNAME} arch)
-		pkg -o IGNORE_OSVERSION=yes -o ABI="FreeBSD:${VERSION}:${ARCH}" -o REPOS_DIR="${JAILMNT}/etc/pkg" -r "${JAILMNT}" update || \
+		pkg -o IGNORE_OSVERSION=yes -o ABI="FreeBSD:${VERSION}:${ARCH#*.}" -o REPOS_DIR="${JAILMNT}/etc/pkg" -r "${JAILMNT}" update || \
 			err 1 "pkg update failed"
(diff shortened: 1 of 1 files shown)


jlduran:
Test plan:

```console
# poudriere jail -c -j generic-15 -m pkgbase=base_latest -U https://pkg.freebsd.org -v 15 -K generic
# poudriere jail -l
JAILNAME    VERSION         OSVERSION ARCH          METHOD  TIMESTAMP           PATH
generic-15  15.0-CURRENT              arm64.aarch64 pkgbase 2024-10-15 16:27:01 /usr/local/poudriere/jails/generic-15
```

jlduran:
cc/ @emaste

jlduran:
> [!NOTE]
> The `OSVERSION` does not show up because there is no `/sys` in the jail (not related to this change).

bdrewery:
I don't use pkgbase or touch image ever, so need a review from someone else to get this in.

jlduran:
Rebase.

Let's wait for @bapt or @manu then. Thank you!

I'll take the rebase as an opportunity to explain the reasoning behind the changes.

This pull request contains two commits:

1. `get_host_arch()` sets `ARCH` as `$(uname -m).$(uname -p)`, if both are the same (as is the case for amd64), then it just uses one. For all the other architectures, both values will differ.  `pkg` ABI uses `FreeBSD:<version>:$(uname -p)`, so we must trim the `uname -m` part.

2. On some installations, `make distributeworld` may have not been executed (as is the case for pkgbase).  Among other things, it installs the symlink `/sys ->/usr/src/sys` (https://github.com/freebsd/freebsd-src/blob/c3f8900e696998c410dc16f9bd9d45c24c413e6b/Makefile.inc1#L1522).  Arguably, this symlink can be added to the pkgbase as a post-install scrpt.

dch:
Urgently needed for testing 15.0 pre-release pkgbase :-)

- works against official snapshots
- works for locally built pkgbase too

@jlduran can you rebase this so it patches cleanly again against HEAD?

jlduran:
> Urgently needed for testing 15.0 pre-release pkgbase :-)
> 
> * works against official snapshots
> * works for locally built pkgbase too
> 
> @jlduran can you rebase this so it patches cleanly again against HEAD?

Done!

bdrewery (approved):

bdrewery:
Merging this in. Thank you. Sorry for the long wait.

PR #1194, opened 2024-12-23
image_zfs: use default (fletcher4) checksum algorithm

svmhdvn:
fletcher4 is the default as listed in `zfsprops(7)` and `sha512` doesn't seem to be required unless there's some explicit need for a cryptographic hash.

### DIFF
--- a/src/share/poudriere/image_zfs.sh
+++ b/src/share/poudriere/image_zfs.sh
@@ -67,7 +67,7 @@ zfs_prepare()
 	zpool create \
 		-O mountpoint=/${ZFS_POOL_NAME} \
 		-O canmount=noauto \
-		-O checksum=sha512 \
+		-O checksum=on \
 		-O compression=on \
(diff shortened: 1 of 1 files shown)


bdrewery:
Yeah I noticed this myself too. I'm thinking of making this more customizable. It's so opinionated right now. I think it's intended to match src/release/tools/vmimage.subr `buildfs` but it doesn't completely.

src/share/poudriere/image_zfs.sh
```diff
@@ -67,7 +67,7 @@ zfs_prepare()
 	zpool create \
 		-O mountpoint=/${ZFS_POOL_NAME} \
 		-O canmount=noauto \
-		-O checksum=sha512 \
+		-O checksum=on \
```
bdrewery, line 70: Do we even need to specify `checksum=on`?

bdrewery (approved):

PR #1197, opened 2025-01-02
Introduce JAILED_DATASET option to poudriere.conf

arrowd:
Rationale for the change:

We at $WORK are using Poudriere as a part of the CI process. For a selected set of ports we define `WITH_TESTING_PORTS`, which makes the `test` target a part of the pipeline. It works quite well for many of ports we're using.

However, one of the projects we've developed requires creating a ZFS dataset as part of its tests. The #1175 PR made it possible for `zfs` command to run, but no datasets or pools are visible from within the jail. This follow-up change passes through a throw-away ZFS dataset into the jail and gives full access to it.

I understand that the feature is quite niche, but it doesn't add much complexity to the code, so I hope it will be accepted.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -967,12 +967,19 @@ injail_tty() {
 jstart() {
 	local mpath name network
 	network="${LOCALIPARGS:?}"
+	local allow_mount_args
 
 	case "${RESTRICT_NETWORKING-}" in
 	"yes") ;;
 	*)
 		network="${IPARGS:?} ${JAIL_NET_PARAMS-}"
 		;;
+	esac
+
+	case "${JAILED_DATASET}" in
+	"") ;;
+	*)
+		allow_mount_args="allow.mount=1 allow.mount.zfs=1 enforce_statfs=1"
 	esac
 
 	_my_name name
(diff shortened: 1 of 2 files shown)


bdrewery:
Could we add some hooks, or use existing hooks, to achieve this?

arrowd:
I agree, the hooks feels like a more proper place for this feature, but from what I gather, even the "jail start" hook is ran after the jail is actually created, so there is no way for it to pass `allow.mount=1 allow.mount.zfs=1 enforce_statfs=1`.

arrowd:
How do we move this forward?

arrowd:
Bump. Can we get this in?

bdrewery:
The problem is that you want it _in_ so it can be maintained upstream. But without _tests_ it will quickly break and rot and I can't promise it will work tomorrow. It is very niche and I don't see the use case outside of your special port. I think we need to add more hooks to let you have your code run exactly where and how needed.

> I agree, the hooks feels like a more proper place for this feature, but from what I gather, even the "jail start" hook is ran after the jail is actually created, so there is no way for it to pass allow.mount=1 allow.mount.zfs=1 enforce_statfs=1.
```
allow_mount_args="allow.mount=1 allow.mount.zfs=1 enforce_statfs=1"
```
This could be set in poudriere.conf `JAIL_PARAMS`.

I'll comment in the code where I think to try hooks.

src/share/poudriere/common.sh
```diff
@@ -1035,6 +1050,10 @@ jstop() {
 	_my_name name
 	jail -r "${name:?}" 2>/dev/null || :
 	jail -r "${name:?}-n" 2>/dev/null || :
+
+	if [ "${JAILED_DATASET}" ]; then
+		zfs destroy -Rf ${ZPOOL}${ZROOTFS}${JAILED_DATASET}_${name:?} 2>/dev/null || :
+	fi
```
bdrewery: `run_hook jstop post "${name}"`

bdrewery:
If the hooks work for you let me know and I can add them in or you can PR it. If you need an ability for a hook to _return_ something we can discuss that too.

arrowd:
Thank you. I only rebased this PR to catch up with upstream changes and solve conflicts. This is now a very low priority for me, so maybe you should remove this PR from the `3.5.0` milestone.

arrowd:
If I bump in this issue again, I will try out the suggested hooks approach. For now, let's close this.

PR #1199, opened 2025-01-10
image: fix pkgbase repo location

svmhdvn:
* Replace wrongly hardcoded `local` pkgbase repository path with the actual path used by the jail when created with `-m pkgbase=...`
* Remove unnecessary temporary creation/deletion of the repo conf file and replace with the simpler `pkg -R <repo_path_inside_mnt` flag
* Allow for the creation of a pkgbase image without `PKGBASELIST`. This will install all `FreeBSD-*` packages.
* Simplify `-P PKGBASELIST` logic using `xargs`, which should also handle empty lines correctly now
* Remove redundant `_jget mnt`

### DIFF
--- a/src/share/poudriere/image.sh
+++ b/src/share/poudriere/image.sh
@@ -238,21 +238,18 @@ convert_package_list() {
 
 install_world_from_pkgbase()
 {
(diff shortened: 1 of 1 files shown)


svmhdvn:
Any updates in testing or reviewing this?

Issue #1181, opened 2024-10-21
poudriere-image generate wrong / owner with overlaydir option


ocochard:
## Prerequisites
- [X] Have you checked for an existing issue describing your problem?
- [X] Are you running the latest version?
- [X] Is your ports tree recent?
- [X] Is your FreeBSD Host on a [supported release](https://www.freebsd.org/security/#sup)?

## Describe the bug
<A clear and concise description of what the bug is.>
Generating a poudriere-image with option `-c overlaydir`, when this directory belongs to a local system user, broke permission on the generated filesystem.

## How to reproduce

with a local simple user:
```
mkdir -p /tmp/overlay/root
mkdir -p /tmp/overlay/etc/dummy1.txt
touch /tmp/overlay/root/dummy2.txt
sudo poudriere image ... -c /tmp/overlay
```
The generated image will have all existing directories belonging to an unknown user (because host local user id doesn’t exist on generated image), in this example its /, /etc and /root.
These bug is due to the [`cp -fRpr` used](https://github.com/freebsd/poudriere/blob/3.4.2/src/share/poudriere/image.sh#L525).
Note here:  we could replace `-Rpr` by `-a`.

## Expected behavior
We could not preserve original user owner when copying the overlay directory: There is no local user created by default on the targeted disk image, and an already existing directory should not have this owner overwrited.

ocochard:
A potential replacement of the `cp` could be something like `(cd "${EXTRADIR}" && find . -print0 | sudo cpio -pdm0 --no-preserve-owner ${WRKDIR}/world/)`

jlduran:
What I have been doing to work around this issue is having the same uids/gids on the builder as on the target. This way the mapping is one-to-one. Pushing the overlay directory to a git repository, or provisioning form another builder may indeed be problematic.

However, there are times I need to have a specific uid/gid (`/tmp/overlay/usr/local/etc/raddb`, for example with a uid:gid of `133:133`).  Perhaps, another option could be an mtree-based solution, where the ownership "metadata" is stored.

PR #1200, opened 2025-01-20
Allow to fix ownership and permission on files copied from the overla…

ocochard:
…y directory.

### DIFF
--- a/src/share/poudriere/image.sh
+++ b/src/share/poudriere/image.sh
@@ -58,7 +58,8 @@ Options:
                        the disk image and mount it to \$WRKDIR/world before
                        installing the contents to the image
     -c overlaydir   -- The content of the overlay directory will be copied into
-                       the image
+                       the image. Owners and permissions will be overwritten if
+                       an <overlaydir>.mtree file is found
     -f packagelist  -- List of packages to install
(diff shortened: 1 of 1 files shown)


ocochard:
Related to issue https://github.com/freebsd/poudriere/issues/1181

src/share/poudriere/image.sh
```diff
@@ -58,7 +58,8 @@ Options:
                        the disk image and mount it to \$WRKDIR/world before
                        installing the contents to the image
     -c overlaydir   -- The content of the overlay directory will be copied into
-                       the image
+                       the image. Owners and permissions will be overwritten if
+                       an overlaydir.mtree file is found
```
evadot, line 62: A bit confusing as people might think that the file needs to be strictly named 'overlaydir.mtree'

ocochard: Could using `$overlaydir.mtree` enough to solve this confusion ?

evadot: Or <overlaydir>.mtree

evadot: Stupid github, that was supposed to be \<overlaydir\>.mtree

PR #1204, opened 2025-01-31
Workaround issue with 32bit libraries

madpilot78:
Poudriere is forcing rebuilds each and every time for some big packages (llvm, gcc, rust) lamenting some missing libraries like `libc.so.7:32`.

I have created this patch to work around the issue, although I'm not sure this is the correct fix or if there is some more solid issue laying behind it.

But I hope this can help discover what is wrong, and be used as a temporary band-aid.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -7311,7 +7311,7 @@ package_libdeps_satisfied() {
 		shlib_name="${shlib%.so*}"
 		case " ${shlibs_provided} " in
 		# Success
-		*" ${shlib} "*) ;;
+		*" ${shlib%:32} "*) ;;
 		# A different version! We need to rebuild to use it.
 		# This supports X.Y.Z for each 0-999.
 		# There is probably a better way to do this. We need to see


bapt:
The workaround seems wrong, with pkg 2.X we track 32bit required libraries, but we also track 32bit required libraries. so the live above your change shoud match.

madpilot78:
@bapt I understand, but with debugging information in poudriere I see the provided libraries miss al the 32bit ones on my system.

But maybe this is due to me needing to rebuild all the jails from scratch, which I will try doing soon. A stated in the mailing lists, after your explanation, I'm not sure how to "fix" a jail built from sources.

bdrewery:
What's a specific port this shows up with? I can debug it more.

nunotexbsd:
# poudriere testport -j 15amd64 -p main games/rocksndiamonds
<repeat testport>
# poudriere testport -j 15amd64 -p main games/rocksndiamonds
Warning: llvm19-19.1.7 will be rebuilt as it misses libc++.so.1:32

(With PKG_NO_VERSION_FOR_DEPS=yes)

madpilot78:
Hi again, while the patch fixes the issue with llvm and other ports, some others insist on requiring a rebuild.

Here are the ones I noticed:

- `x11/nvidia-driver-470` (misses `libX11.so.6:32`)
- `lang/go121` (misses `libc.so.6:32`)
- `lang/go123` (misses `libc.so.6:32`)

bapt:
On Wed 12 Feb 23:24, Guido Falsi wrote:
> Hi again, while the patch fixes the issue with llvm and other ports, some others insist on requiring a rebuild.
> 
> Here are the ones I noticed:
> 
> - `x11/nvidia-driver-470` (misses `libX11.so.6:32`)

This one there is nothing poudriere can do about it, as there is nothing which
provides this in the ports tree. this is a defect of the port, why do it install
a 32bit binary which depends on a library which nothing provides: aka real bug
in the ports.

> - `lang/go121` (misses `libc.so.6:32`)
> - `lang/go123` (misses `libc.so.6:32`)

/usr/local/go122/src/debug/elf/testdata/gcc-386-freebsd-exec
this file requires libc.so.6:32 this binary cannot work at all on FreeBSD.

So poudriere is right reporting the issue.

madpilot78:
@bapt Thanks for the analysis!

I guess some fix will be required for these at the port level then.

bdrewery:
Hmm even if those ports are fixed Poudriere only knows how to label base libs as 32bit. It will need to also learn how to label ports libs as 32bit. Feel free to open a bug if there is a real case of this found.

PR #1206, opened 2025-02-05
Update clean snapshot after upgrade using pkgbase

rbgarga:
- **jail.sh: Remove extraneous space**
- **jail.sh: Re-create clean snapshot after upgrade**


bdrewery:
Does this need to happen at the end of `update_pkgbase` too? I'm going to merge this for now.

PR #1210, opened 2025-03-12
auto_lib_depends: stop considering LIB_DEPENDS as RUN_DEPENDS

bapt:
if the ports tree have the AUTO_LIB_DEPENDS feature flag, then it means it stops registring the LIB_DEPENDS as RUN_DEPENDS to only rely on shlibs_required/shlibs_provides which means poudriere has to stop analysing for LIB_DEPENDS to check if they changed between the already compiled version of the packages and the current port.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -6936,7 +6936,14 @@ _delete_old_pkg() {
 		# which will avoida all of the injail hacks
 
 		# pkgname-lib_deps pkgname-run_deps
-		for td in lib run; do
+		dep_types=""
+		if have_ports_feature AUTO_LIB_DEPENDS; then
+			dep_types="run"
+		else
+			dep_types="lib run"
+		fi
+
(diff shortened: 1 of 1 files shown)


bdrewery:
https://github.com/freebsd/poudriere/tree/bdrewery/auto_lib_depends has the rest of the changes needed for this.
Some discussion in https://github.com/freebsd/poudriere/blob/bdrewery/auto_lib_depends/docs/shlib_tracking.md.
I have not given enough thought to the current incremental build that this PR specifically is modifying. My intuition suggests we may miss rebuilding some packages when `LIB_DEPENDS` changes. If that's the case I need to revert 6471b02b20eb0e2bc21f51ec600514d029f33031 to enable the deeper inspection that `PKG_NO_VERSION_FOR_DEPS` enables.

bdrewery:
Merged in 67daecb2cfdc70917a278af73840819d7f569dc4

PR #1211, opened 2025-03-13
Remove undeleted file in .gitignore

fel1x-developer:
Removed with

```sh
$ grep -v '^#' .gitignore | xargs -I {} git rm -r --cached "{}"
```

Related: #1108


bdrewery:
This is a critical file. Proper fix is to remove it from .gitignore. Done in d37d416. Thank you.

PR #1218, opened 2025-04-07
fetch: only update FreeBSD repo

svmhdvn:
This fixes an issue where a fetch fails because a builder (pkgbase) jail contains multiple repo entries (one for FreeBSD and one for pkgbase at least), but the pkgbase repo isn't mounted in the jail.


linnemannr:
This actually breaks -b with poudriere-bulk, because the pkg rquery returns nonzero if any of the enabled repo dbs are not initialized, for example FreeBSD-kmod. You can work around this by modifying the /etc/pkg/FreeBSD.conf in the jail root before running it, but that's pretty filthy.

PR #1229, opened 2025-06-17
conf: set PKG_REPRODUCIBLE=yes by default

svmhdvn:
`PKG_REPRODUCIBLE` is a sane default for poudriere and is currently not discoverable in the `poudriere.conf.sample` file. Tested along with https://reviews.freebsd.org/D50911.


bdrewery:
Great idea. Shouldn't it be default on in ports for `PACKAGE_BUILDING`?

svmhdvn:
Sorry, I didn't understand what you meant. This should be a trivial addition to the conf file right?

bdrewery:
Ah I see. I had forgotten this was a Poudriere thing. It's a good idea.
There is a top-level git hash that is inserted as metadata that will break this rule but I can fix that separately.

bdrewery:
Although I'm indecisive on how `PKG_REPRODUCIBLE` should behave with the top-level git hash. Should we say it is reproducible from the same ports top-level checkout or same port-level checkout? I think it makes sense at top-level since that contains so much framework that can change a port output. So in that sense `PKG_REPRODUCIBLE` shouldn't remove the top-level hash.
Even with all the reproducibility support added in, the same `PORTVERSION` could have 2 different packages if the framework changes (or a dependency with the current default incremental build algorithm).
 But it's still going to be noisy and surprise the purpose of the flag.

bdrewery:
@svmhdvn What are your thoughts on this? I had to remove the port-level git hash metadata as it was too expensive. I could explore bringing it back if it helps here.

svmhdvn:
> I think it makes sense at top-level since that contains so much framework that can change a port output.

Agreed, definitely it should be done on a whole ports tree level checkout.

> But it's still going to be noisy and surprise the purpose of the flag.

Noisy is generally good IMO, as any "false positive" is better than missing important details changed by the ports framework. Since reproducible builds is all about creating the same output from the same input (including the toolchain and environment), this should be ok. Someone else verifying reproducibility will have to have the same ports checkout anyway.

bdrewery:
Cool. Thanks for taking this.

PR #1241, opened 2025-08-11
Add overlays to the jail command

oshogbo:
I think there is a missing option for running a reference jail when working with overlays.
We have all the required utilities. We just need to pass the information about the overlays.


vermaden:
The only reason these checks failed is: **Monthly compute limit exceeded!**

vermaden:
What is blocking adding this?

@oshogbo Sent a ready to apply patch for a needed and wanted feature ...

PR #1245, opened 2025-09-07
jail: ensure ALPHA builds are picked up from /releases/ url

dch:
The alpha builds are published at /releases/ not /snapshots/ so
the fetch fails as it believes ALPHA should be in snapshots.

e.g. https://download.freebsd.org/ftp/releases/arm64/15.0-ALPHA1/

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -833,7 +833,7 @@ install_from_ftp() {
 		case ${METHOD} in
 			ftp|http|gjb)
 				case ${VERSION} in
-					*-CURRENT|*-ALPHA*|*-PRERELEASE|*-STABLE) type=snapshots ;;
+					*-CURRENT|*-PRERELEASE|*-STABLE) type=snapshots ;;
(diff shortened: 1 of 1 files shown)


jlduran:
ALPHA builds should be under snapshots:

https://download.freebsd.org/snapshots/arm64/aarch64/15.0-ALPHA1/

~~What exactly is the issue?~~
EDIT: For some reason ALPHA builds are under releases.

dch:
There’s now a softlink between these locations I’ll check what we are supposed to be doing in future

jlduran:
For example, https://download.freebsd.org/snapshots/ISO-IMAGES/15.0/ should also contain the ISOs, currently in https://download.freebsd.org/releases/ISO-IMAGES/15.0/

Regardless, I do not think this is a Poudriere bug.

jlduran:
This has been fixed for ALPHA2 (maybe freebsd/freebsd-src@1bc7ee74b6a80081f0f5764970e3949e834965f1?)

dch:
indeed

PR #1261, opened 2025-10-07
common/update_version_env(): don't set both ABI_FILE and OSVERSION

svmhdvn:
This fixes an annoying message that shows up for every built port when building with QEMU:

"pkg-static: Both ABI_FILE and OSVERSION are set, ABI_FILE overrides OSVERSION"

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -3465,11 +3465,13 @@ update_version_env() {
 	local osversion="$5"
 	local login_env
 
-	login_env=",UNAME_r=${version% *},UNAME_v=FreeBSD ${version},OSVERSION=${osversion}"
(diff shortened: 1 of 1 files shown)


bdrewery:
I'm not sure about this one. `pkg` owns `ABI_FILE` but it doesn't own `OSVERSION`. `OSVERSION` is used by ports.

bdrewery:
@bapt how should we make this warning go away? The variables are different purposes, why is `pkg` warning about it?

bapt:
if ABI_FILE is defined with a valid version I should probably ignore OSVERSION if defined in env and not via command line and bot issue warning in this case

bapt:
aka yes this should fixed on pkg side

PR #1265, opened 2025-10-11
Fix pkgclean respecting PACKAGES being set in a jail-poudriere.comf

brd:


bdrewery:
I think we are missing tests for `poudriere audit`. I'll write one up and send it to you for review.
There are `poudriere pkgclean` ones but they are temporarily disabled for CI while I reduce the CPU they use.

bdrewery:
The test PR I mentioned won't be here soon. No need to wait for it.

brd:
Thanks for the review, will merge soon, was just doing some extra verification.

PR #1269, opened 2025-10-14
Add GIT_PROXY option to use Poudriere behind proxy w/o direct Internet connection

vermaden:
Hi,

I added **GIT_PROXY** option to **poudriere.conf.sample** and also 'handling' of that option in **jail.sh** and **ports.sh** files.

Worked well on my system.

Let me know If I did not miss anything.

Regards,
vermaden

### DIFF
--- a/src/share/poudriere/ports.sh
+++ b/src/share/poudriere/ports.sh
@@ -297,6 +297,9 @@ create)
 			if [ ! -x "${GIT_CMD}" ]; then
 				err 1 "Git is not installed. Perhaps you need to 'pkg install git'"
 			fi
+			msg_n "Setting up proxy..."
+			${GIT_CMD} config --system http.proxy ${GIT_PROXY} || err 1 " fail"
+			echo " done"
 			msg_n "Cloning the ports tree..."
 			${GIT_CMD} clone ${GIT_DEPTH} ${quiet} \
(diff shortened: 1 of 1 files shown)


vermaden:
Ok, this patch is not needed at all ... its just `git config --global` is useless with Poudriere (even executed as root) but `git config --system` makes Poudriere work properly.

Example:

```
# git config --global http.proxy http://proxy.freebsd.xyz:3128

# poudriere ports -c
[00:00:00] Creating default fs at /var/local/poudriere/ports/default... done
[00:00:00] Cloning the ports tree...fatal: unable to access 'https://git.FreeBSD.org/ports.git/': Could not resolve host: git.FreeBSD.org
[00:00:00] Error: /usr/local/share/poudriere/ports.sh:303: fail
[00:00:00] Error while creating ports tree, cleaning up.

# git config --system http.proxy http://proxy.freebsd.xyz:3128

# poudriere ports -c                                               
[00:00:00] Creating default fs at /var/local/poudriere/ports/default... done
[00:00:00] Cloning the ports tree...
```

Regards,
vermaden

vermaden:
I can of course rework the patch with `git -c http.proxy=${GIT_PROXY} rest_of_git_command` if someone finds that useful.

PR #1305, opened 2026-02-10
jail, ports: support subcommand style arguments

nanorkyo:
Allow users to use subcommands (e.g., 'poudriere jail create') as an alternative to the traditional flag-based arguments (e.g., 'poudriere jail -c'). This improves CLI usability and intuitiveness.

Supported subcommands:
- jail: create(-c), delete(-d), list(-l), update(-u), info(-i), start(-s), stop(-k)
- ports: create(-c), delete(-d), list(-l), update(-u)

Backward compatibility with existing flags is strictly maintained.

### DIFF
--- a/src/man/poudriere-jail.8
+++ b/src/man/poudriere-jail.8
@@ -36,6 +36,9 @@
 .Nd manage jails used to build ports
 .Sh SYNOPSIS
 .Nm
+.Cm subcommand
+.Op Ar options
+.Nm
 .Fl c
 .Fl j Ar name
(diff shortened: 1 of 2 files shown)


bdrewery:
This is cool. Could you update the manpages and shell completions as well?
We should also make `set_command` fail if ran a second time.

nanorkyo:
Oops,  I forgot them.  I'll update the manpages and shell completions.

> We should also make set_command fail if ran a second time.

OK, I'll fix this behavior.

PR #1313, opened 2026-02-27
logclean: Fix empty build log directory removal with multiple jails

### LINKED ISSUE #1157: logclean: Fix cleanup for multiple MASTERNAMEs
## Describe the bug
When multiple jails (on my system there were `neptune-amd64`, `saturn-amd64` and `sirius-amd64`) are used, the first `logclean` run would fail at "Removing empty build log directories" step with, e.g.:

```
[00:00:00] Removing empty build log directories...find: neptune-amd64-default saturn-amd64-default sirius-amd64-default/latest-per-pkg: No such file or directory
Error: (77076) /usr/local/share/poudriere/logclean.sh:422: set -e error: status = 1
```

## How to reproduce
1. Use poudriere to generate logs, this can be done by using multiple jails to perform an empty `bulk` build.
2. Run `poudriere logclean -ay`

## Expected behavior
The whole process should complete without error.

## Environment
 - Host OS: 14.1 amd64
 - Jail OS: various, but mainly 14.1 amd64 and 15.0 amd64
 - Browser: N/A
 - Poudriere Version: poudriere-devel-3.4.99.20240424 from ports tree
 - Ports branch and revision: main

## Additional context

I would propose the following change:

```
diff --git a/src/share/poudriere/logclean.sh b/src/share/poudriere/logclean.sh
index f28ecc44..99bcf920 100755
--- a/src/share/poudriere/logclean.sh
+++ b/src/share/poudriere/logclean.sh
@@ -370,11 +370,11 @@ if [ ${logs_deleted} -eq 1 ]; then
 
        msg_n "Removing empty build log directories..."
        if lock_have "logs_latest-per-pkg"; then
-               echo "${MASTERNAMES_LOCKED:?}" | sed -e 's,$,/latest-per-pkg,' | \
-                   tr '\n' '\000' | \
-                   xargs -0 -J % find -x % -mindepth 0 -maxdepth 0 -empty | \
+               for MASTERNAME in ${MASTERNAMES_LOCKED:?}; do
+                   find -x "${MASTERNAME}/latest-per-pkg" -mindepth 0 -maxdepth 0 -empty | \
                    sed -e 's,$,/..,' | xargs realpath | tr '\n' '\000' | \
                    xargs -0 rm -rf
+               done
                echo " done"
        else
                echo " skipped (locked by another process)"
```

Basically, `MASTERNAMES_LOCKED` was a set of strings separated by blank space.  For *all* other usages, `for MASTERNAME in ${MASTERNAMES_LOCKED:?};` idiom was used, and for this one case the sed seems to be both redundant and incorrect (because it adds `/latest-per-pkg` to the end of the _list_, and the code seems to be expecting the list to be `\n` separated and it is not).

I _think_ it makes more sense to simply avoid these `sed`'s and manipulate the string in shell.

delphij:
When multiple jails are configured, the "Removing empty build log directories" step fails because the script incorrectly processes the space-separated MASTERNAMES_LOCKED list.

The issue is that sed appends "/latest-per-pkg" to the end of the entire space-separated string rather than to each jail name individually. This results in a malformed path like:
  "neptune-amd64-default odin-amd64-default saturn-amd64-default/latest-per-pkg"

Fix by converting spaces to newlines before the sed operation, ensuring each jail name gets the suffix appended correctly.

Fixes #1157

### DIFF
--- a/src/share/poudriere/logclean.sh
+++ b/src/share/poudriere/logclean.sh
@@ -398,7 +398,7 @@ if [ ${logs_deleted} -eq 1 ]; then
 	if lock_have "logs_latest-per-pkg"; then
 		case "${MASTERNAMES_LOCKED:+set}" in
 		set)
-			echo "${MASTERNAMES_LOCKED:?}" | sed -e 's,$,/latest-per-pkg,' | \
+			echo "${MASTERNAMES_LOCKED:?}" | tr ' ' '\n' | sed -e 's,$,/latest-per-pkg,' | \
 			    tr '\n' '\000' | \
 			    xargs -0 -J % find -x % -mindepth 0 -maxdepth 0 -empty | \
 			    sed -e 's,$,/..,' | xargs realpath | tr '\n' '\000' | \


bdrewery:
Thanks for submitting the PR!

PR #1323, opened 2026-04-08
Fix "bulk -b"'s pkg version check with certain package servers

asomers:
When using "bulk -b" or "testport -b", Poudriere will check that the remote server's pkg version is less than or equal to the local package version before fetching any remote packages.  That process entails running "pkg rquery -U" from within the build jail.  But in some setups pkg-rquery will return two results: one for the remote server and one for the local package cache.  That screws up the shell script's parsing of the output.  I don't know why this doesn't happen to everybody, but the solution is simple: use rquery's "-r" option.

Sponsored by:	ConnectWise

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -4806,7 +4806,7 @@ download_from_repo() {
 		return 0
 	fi
 
-	remote_pkg_ver="$(injail ${pkg_bin} rquery -U %v "${P_PKG_PKGBASE:?}")"
+	remote_pkg_ver="$(injail ${pkg_bin} rquery -U -r "${PACKAGE_FETCH_BRANCH}" %v "${P_PKG_PKGBASE:?}")"
 	local_pkg_name="${P_PKG_PKGNAME:?}"
 	local_pkg_ver="${local_pkg_name##*-}"
 	case "$(pkg_version -t "${remote_pkg_ver}" "${local_pkg_ver}")" in


src/share/poudriere/common.sh
bdrewery, line 4791: If I understand what you're addressing right, this is what is supposed to prevent the problem. Do you have a custom repo that is not being disabled? I thought `-r` made sense too but the solution converged on this route. More research is needed to understand why `-r` was rejected previously.

bdrewery: It could simply be that `-r` was invasive.
Somewhere in https://github.com/freebsd/poudriere/issues?q=label%3A%22Package%20Fetch%22 (and linked PRs) lies the issue. An agent could read through these quickly.

asomers: Yes I have a custom repo that isn't being disabled by those lines.  Is the stuff in /packages supposed to represent what was fetched from the remote server?  If so, then my change is wrong and I should be disabling our custom repo instead.
FWIW it looks like you are the one who originally added the code to disable remote repos, way back in 2014: f82f7e13e918bc8fc658104a197753967a7a20de .  That's two years after you also added -r to `pkg rquery`.  And the check for the pkg version was added in 2021 c88db0801ed2937bda9a170405a59ab3f8897938, well after the decision to disable remote repos.
I guess my problem is that I don't understand the role of /packages .  Should we be using "-r local" to _only_ query from /packages, and call that the remote package version?

bdrewery: Well this is just about the upstream fetching so I'd think we'd want to ignore your custom repo. Is that right?

/packages is the existing package set.

Adding `-r Poudriere` to _every_  `${pkg_bin}` call in `download_from_repo` would be needed to limit it to the `PACKAGESITE` being passed in.

Something like this, which I can't test at the moment.
```diff
diff --git src/share/poudriere/common.sh src/share/poudriere/common.sh
index 1dc773824..7c9b931e6 100755
--- src/share/poudriere/common.sh
+++ src/share/poudriere/common.sh
@@ -4807,19 +4807,19 @@ download_from_repo() {
 	# Bootstrapping might occur here.
 	# XXX: rquery is supposed to 'update' but it does not on first run.
 	if ! JNETNAME="n" injail env ASSUME_ALWAYS_YES=yes \
-	    ${pkg_bin} update -f; then
+	    ${pkg_bin} update -r Poudriere -f; then
 		msg "Package fetch: Not fetching as remote repository is unavailable."
 		rm -f "${missing_pkgs}"
 		return 0
 	fi
 	# Don't trust pkg-update to return its error
-	if ! injail ${pkg_bin} rquery -U %n pkg >/dev/null; then
+	if ! injail ${pkg_bin} rquery -r Poudriere -U %n pkg >/dev/null; then
 		msg "Package fetch: Failed to fetch package repository."
 		rm -f "${missing_pkgs}"
 		return 0
 	fi
 
-	remote_pkg_ver="$(injail ${pkg_bin} rquery -U %v "${P_PKG_PKGBASE:?}")"
+	remote_pkg_ver="$(injail ${pkg_bin} rquery -r Poudriere -U %v "${P_PKG_PKGBASE:?}")"
 	local_pkg_name="${P_PKG_PKGNAME:?}"
 	local_pkg_ver="${local_pkg_name##*-}"
 	case "$(pkg_version -t "${remote_pkg_ver}" "${local_pkg_ver}")" in
@@ -4834,25 +4834,25 @@ download_from_repo() {
 	# (like pkg rquery -U), and it uses various locking that isn't needed
 	# here. Grab all the options for comparison.
 	remote_all_options=$(mktemp -t remote_all_options)
-	injail ${pkg_bin} rquery -U '%n %Ok %Ov' > "${remote_all_options}"
+	injail ${pkg_bin} rquery -r Poudriere -U '%n %Ok %Ov' > "${remote_all_options}"
 	remote_all_pkgs=$(mktemp -t remote_all_pkgs)
-	injail ${pkg_bin} rquery -U '%n %n-%v %?O' > "${remote_all_pkgs}"
+	injail ${pkg_bin} rquery -r Poudriere -U '%n %n-%v %?O' > "${remote_all_pkgs}"
 	remote_all_deps=$(mktemp -t remote_all_deps)
-	injail ${pkg_bin} rquery -U '%n %dn-%dv' > "${remote_all_deps}"
+	injail ${pkg_bin} rquery -r Poudriere -U '%n %dn-%dv' > "${remote_all_deps}"
 	remote_all_annotations=$(mktemp -t remote_all_annotations)
 	remote_all_cats=$(mktemp -t remote_all_cats)
 	case "${IGNORE_OSVERSION-}" in
 	"yes") ;;
 	*)
-		injail ${pkg_bin} rquery -U '%n %At %Av' > "${remote_all_annotations}"
-		injail ${pkg_bin} rquery -U '%n %C' > "${remote_all_cats}"
+		injail ${pkg_bin} rquery -r Poudriere -U '%n %At %Av' > "${remote_all_annotations}"
+		injail ${pkg_bin} rquery -r Poudriere -U '%n %C' > "${remote_all_cats}"
 		;;
 	esac
 	abi="$(injail "${pkg_bin}" config ABI)"
 	remote_all_abi=$(mktemp -t remote_all_abi)
-	injail ${pkg_bin} rquery -U '%n %q' > "${remote_all_abi}"
+	injail ${pkg_bin} rquery -r Poudriere -U '%n %q' > "${remote_all_abi}"
 	remote_all_prefix=$(mktemp -t remote_all_prefix)
-	injail ${pkg_bin} rquery -U '%n %p' > "${remote_all_prefix}"
+	injail ${pkg_bin} rquery -r Poudriere -U '%n %p' > "${remote_all_prefix}"
 
 	parallel_start || err 1 "parallel_start"
 	wantedpkgs=$(mktemp -t wantedpkgs)
@@ -4878,6 +4878,7 @@ download_from_repo() {
 		return
 	fi
 
+	# XXX: this is wrong
 	packagesite_resolved=$(injail ${pkg_bin} -vv | \
 	    awk '/[[:space:]]*url[[:space:]]*:[[:space:]]*/ {
 		    gsub(/^"|",$|,$/, "", $3)
@@ -4896,7 +4897,7 @@ download_from_repo() {
 	    err 1 "null mount failed for pkg cache"
 	if ! JNETNAME="n" injail xargs \
 	    env ASSUME_ALWAYS_YES=yes \
-	    ${pkg_bin} fetch -U < "${wantedpkgs}"; then
+	    ${pkg_bin} fetch -r Poudriere -U < "${wantedpkgs}"; then
 		msg "Package fetch: Error fetching packages"
 		umountfs "${MASTERMNT:?}/var/cache/pkg"
 		rm -f "${wantedpkgs}"
```

asomers: My custom repo and Poudriere end up being identical.  They both refer to the same server.  And your change works for me.

bdrewery: I'm going to close this and open an issue to track adding the `-r` flag.

bdrewery: Moved to #1325

PR #1326, opened 2026-04-14
rquery: fix usage to be compatible with all pkg versions

bapt:
sorry I haven't see the rquery bug was used in poudriere, this should fix the usage

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -4821,17 +4821,17 @@ download_from_repo() {
 	# (like pkg rquery -U), and it uses various locking that isn't needed
 	# here. Grab all the options for comparison.
 	remote_all_options=$(mktemp -t remote_all_options)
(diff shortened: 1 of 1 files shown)


bdrewery:
@bapt it's not working or there are other problems. Is there a way to make it less strict again?
```
pkg-static: mailest: duplicate dependency listing: mew
unexpected end of expression
Error: (8178) /usr/local/share/poudriere/testport.sh:301:prepare_ports:10438:download_from_repo:4847:injail:997:injail_direct:1024: set -e error: status = 1
```

```
# sed -n 4847p /usr/local/share/poudriere/common.sh
                injail ${pkg_bin} rquery -U -e '%#A' '%n %At %Av' > "${remote_all_annotations}"
```

bdrewery:
Also what about this one?
```
src/share/poudriere/image.sh:   pkg rquery '%At %o@%Av %n-%v' | \
```
Not clear to me where we need `-e` now.

PR #1327, opened 2026-04-14
fix fallout from the name-(null) which is now name-

bapt:
another fallout sorry again


bdrewery:
Aha thanks for fixing the hack poudriere is using!

clausecker:
Works for me.

bdrewery:
poudriere-devel is updated in main now

PR #1334, opened 2026-04-21
testport: do not check the parent directory of a port does not have Mk

glebius:
The check originates from d72e70f and is not a comprehensive check that directory is a valid port directory anyway.  But it breaks a custom port tree (built on top of FreeBSD ports) that has some internal Mk directories.

Fixes: #1319


bdrewery:
https://github.com/freebsd/poudriere/commit/06592c4aeeb723eceb599cf9379a9cb75330ce1e has the commit for `release-3.4`.

I won't get a release out for a while. At least your local patch won't conflict. The surrounding code isn't expected to change soon.

PR #1342, opened 2026-05-05
feat: write ports tree git commit hash to .portscommit in packages dir

### LINKED ISSUE #1301: Put the used ports repo hash into data.pkg/data
Hey Bryan,

as discussed on IRC, I'd like to have information about what latest ports repo commit ID were the binaries in the pkg database built from.

The reason for it is, I could set my local poudriere ports repo to that commit, thus avoiding having to rebuild packages with huge resource hogging compilations, that are dependencies of packages that I build with it.

For clarification, here's a recurring example that is now happening again: rust. It currently has `1.92.0` in the binaries and `1.93.0` in the ports tree, which makes poudriere rebuilding it (with the aforementioned resource hogging), only to use it to compile clamd, that depends on it.

`data.pkg/data` could have an extra key with `latest_ports_commit` or something along these lines that I could extract, and then set my local ports repo to it. That way I could avoid going my production server crazy. :)

Thanks in advance!
Regards,
László

karolyi:
After a bulk build, write the top-level ports tree git hash to ${PACKAGES}/.portscommit alongside the existing .buildname and .jailversion files. This allows users to check out their local ports tree to the exact commit the published packages were built from, avoiding unnecessary rebuilds of unchanged ports (e.g. lang/rust).

The atomic repository commit machinery in commit_packages() is updated to symlink and auto-fix .portscommit the same way it handles .buildname and .jailversion.

TODO: pkg needs to be updated to expose this value inside data.pkg / packagesite.pkg so it is available without a separate out-of-band file.

closes #1301

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -3352,13 +3352,13 @@ commit_packages() {
 	# mostly incase pkg adds a new top-level repo or the ports framework
 	# starts creating a new directory
 	find "${PACKAGES:?}/" -mindepth 1 -maxdepth 1 \
-	    \( ! -name '.*' -o -name '.jailversion' -o -name '.buildname' \) |
+	    \( ! -name '.*' -o -name '.jailversion' -o -name '.buildname' -o -name '.portscommit' \) |
 	    while mapfile_read_loop_redir path; do
 		name=${path##*/}
 		[ ! -L "${PACKAGES_ROOT:?}/${name:?}" ] || continue
 		if [ -e "${PACKAGES_ROOT:?}/${name:?}" ]; then
 			case "${name}" in
-			.buildname|.jailversion|\
+			.buildname|.jailversion|.portscommit|\
 			"data.${PKG_EXT}"|data.txz|\
 			"meta.${PKG_EXT}"|meta.txz|\
 			"digests.${PKG_EXT}"|digests.txz|\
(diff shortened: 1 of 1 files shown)


bdrewery (approved):
Looks good. I need to find time to test it myself before merge.

karolyi:
Yay \o/ :)

Whenever time allows, however I'd like to have this merged soon so I could start making the changes in pkg itself.

bdrewery:
> Yay \o/ :)
> 
> Whenever time allows, however I'd like to have this merged soon so I could start making the changes in pkg itself.

Have you gotten buy-in from @bapt?

karolyi:
No, not yet, but I don't think there should be much of a pushback when I'll issue a PR against its repo. :) I don't see any other way this problem could be solved, and it's only an extra field in the JSON data.

Once this merges, I'll consult him too of course.

karolyi:
Hey,

not being pushy, but I'd like to encourage you to have this PR reviewed and merged if you don't have any objections against it. With the current ports HEAD and pkg stand, poudriere wants to recompile rust and ruby. The latter of which wouldn't be an issue for me, but I'm not willing to have my production machine burn its CPU for half an hour with rust. :)

Thanks in advance.

karolyi:
The same thing happens now with cmake-core ... and rust.

karolyi:
friendly bump :)

karolyi:
please don't forget about this one.

Issue #1242, opened 2025-08-26
Can’t create an aarm64 jail using pkgbase: Unknown architecture 'arm64.aarch64' in ABI string


ocochard:
So I’m using a script to build x86 and ARM packages, and trying to switch to pkgbase.
There are 2 problems:

- The command works on x86, but fails on ARM
- There is no exit code, and the jail is empty (with only pkg repo configured)

```
$ ls /mnt/usr/local/pkgbase-repo/FreeBSD\:15\:aarch64/
15.snap20250826171222 latest
$ sudo poudriere jail -c -j generator -v 15 -m pkgbase=latest -U file:///mnt/usr/local/pkgbase-repo/
[00:00:00] Creating generator fs at /usr/local/poudriere/jails/generator... done
[00:00:00] Installing 15 arm64.aarch64 from file:///mnt/usr/local/pkgbase-repo/ ...pkg: Unknown architecture 'arm64.aarch64' in ABI string
pkg: Cannot parse configuration file!
pkg: Unknown architecture 'arm64.aarch64' in ABI string
pkg: Cannot parse configuration file!
pkg: Unknown architecture 'arm64.aarch64' in ABI string
pkg: Cannot parse configuration file!
[00:00:00] Recording filesystem state for clean... done
[00:00:00] Jail generator 15 arm64.aarch64 is ready to be used
```

madpilot78:
I'm seeing similar behaviour while trying to setup poudriere on an arm64 machine, so it happens also if no `-a` option is specified.

jlduran:
Related to #1180?

markmi:
[I will note here that pkg conventions are also involved here, not just poudriere ones.]

FYI: `poudriere jail -c -jgenerator -aaarch64 -U https://pkg.freebsd.org -mpkgbase=base_latest -v 15` works by finding: `https://pkg.freebsd.org/FreeBSD:15:aarch64/base_latest/`

The "aarch64" is literally substituted into the URL used. Also, as far as ABI strings go, FreeBSD:15:arm64.aarch64 is not used anywhere that I'm aware of --but FreeBSD:15:aarch64 notation is the preexisting ABI triple's notation for such.

So `poudriere jail -c -jgenerator -aaarch64 -U file:///mnt/usr/local/pkgbase-repo/ -mpkgbase=base_latest -v 15` would find: `file:///mnt/usr/local/pkgbase-repo/FreeBSD:15:aarch64/base_latest/`

[...]

PR #1349, opened 2026-05-19
jail: explicitly fail if FreeBSD-src-sys can't be installed

dch:
I'm not sure of the purpose of the intricate previous code:

```
pkg -o IGNORE_OSVERSION=yes -o REPOS_DIR="${JAILMNT}/etc/pkg" -o ABI="FreeBSD:${VERSION}:${ARCH#*.}" -r ${JAILMNT}/ search -q '^FreeBSD-src-sys' \
| xargs pkg -o REPOS_DIR="${JAILMNT}/etc/pkg" -r ${JAILMNT}/ install -y
```

But if the repo URL is incorrect, (e.g. points to a path that doesn't have packages) then this fails silently, and you end up with a poudriere jail without FreeBSD-src-sys.

Now it will fail loudly:

```# poudriere jail -c -j bad_pkgbase -v 16 -m pkgbase=latest -U https://pkg.freebsd.org/
[00:00:00] Creating bad_pkgbase fs at /usr/local/poudriere/jails/bad_pkgbase... done
[00:00:00] Installing 16 amd64 from https://pkg.freebsd.org/ ...pkg: Setting ABI requires setting OSVERSION, guessing the OSVERSION as: 1600000
Updating pkgbase repository catalogue...
Fetching meta.conf: 100%     179 B   0.2 kB/s    00:01
Fetching data: 100%    11 MiB   2.8 MB/s    00:04
Processing entries: 100%
pkgbase repository update completed. 36598 packages processed.
All repositories are up to date.
pkg: Setting ABI requires setting OSVERSION, guessing the OSVERSION as: 1600000
pkg: Setting ABI requires setting OSVERSION, guessing the OSVERSION as: 1600000
Updating pkgbase repository catalogue...
pkgbase repository is up to date.
All repositories are up to date.
pkg: No packages available to install matching 'FreeBSD-src-sys' have been found in the repositories
[00:00:08] Error: [26556] /usr/local/share/poudriere/jail.sh:install_from_pkgbase:40:Failed to install FreeBSD-src-sys (needed for sys/param.h)
[00:00:08] Error while creating jail, cleaning up.
[00:00:08] Removing bad_pkgbase jail... done
Exiting with status 1
```

cc @jlduran @evadot who touched this area last

Are there circumstances where this silent failure is OK? Do we have vendors who build `MadRouterCo-src-sys` packages?

I would like to update the manpage to point to where these URLs are documented, but I only find the pkgbase wiki page.

### DIFF
--- a/src/share/poudriere/jail.sh
+++ b/src/share/poudriere/jail.sh
@@ -1039,7 +1039,8 @@ EOF
 	    err 1 "pkg update failed"
 	# Omit the man/debug/kernel/src and tests packages, unneeded for us.
 	pkg -o IGNORE_OSVERSION=yes -o REPOS_DIR="${JAILMNT}/etc/pkg" -o ABI="FreeBSD:${VERSION}:${ARCH#*.}" -r ${JAILMNT}/ search -qCx '^FreeBSD-.*' | grep -vE -- '-man|-dbg|-kernel-|-tests|-src-' | xargs pkg -o REPOS_DIR="${JAILMNT}/etc/pkg" -r ${JAILMNT}/ install -y
-	pkg -o IGNORE_OSVERSION=yes -o REPOS_DIR="${JAILMNT}/etc/pkg" -o ABI="FreeBSD:${VERSION}:${ARCH#*.}" -r ${JAILMNT}/ search -q '^FreeBSD-src-sys' | xargs pkg -o REPOS_DIR="${JAILMNT}/etc/pkg" -r ${JAILMNT}/ install -y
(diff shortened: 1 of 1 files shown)


jlduran:
I think this is fine. It is in the same vein as #1242. In fact, it may be possible that all those `pkg install` calls require an abrupt failure.

> I'm not sure of the purpose of the intricate previous code

Eventually a `pkg_cmd` could emerge with all those flags to clear out the clutter.

> Are there circumstances where this silent failure is OK?

I don't think so. `poudriere-jail` needs the sources (with distsets, it pulls src.txz as well; if it can't, it fails).

> Do we have vendors who build MadRouterCo-src-sys packages?

Probably not yet with Poudriere, but this is something I would like Poudriere to support.

jlduran (approved):

PR #1358, opened 2026-06-01
Add option (-e) to bulk and testport to run port tests

yurivict:
Add a -e flag to 'poudriere bulk' and 'poudriere testport' that runs 'make test-depends' and 'make test' for each port after the build and stage phases complete.

This helps catch missing TEST_DEPENDS by running port tests in a clean environment where only declared dependencies are installed.

Test failures are fatal by default; use -k to make them non-fatal (sets testfailure=2 and continues the build).

When combining -e with -t (PORTTESTING), the prestage filesystem snapshot is re-recorded after test-depends installs packages so that check_fs_violation at the package phase does not flag test dependencies as staging violations.

### DIFF
--- a/src/man/poudriere-bulk.8
+++ b/src/man/poudriere-bulk.8
@@ -38,7 +38,7 @@
 .Nm
 .Fl a
 .Fl j Ar name
-.Op Fl CcFHIikNnRrSTtvw
+.Op Fl CceFHIikNnRrSTtvw
 .Op Fl B Ar name
 .Op Fl b Ar branch
 .Op Fl J Ar maxjobs Ns Op Cm \&: Ns Ar prebuildmaxjobs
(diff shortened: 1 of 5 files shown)


arrowd:
Note that there is `Mk/Features/testing.mk` that seemingly does the same thing on the framework level.

bdrewery:
This isn't as simple as it seems. This may work for _some_ ports but as soon as it is committed people will open bugs against individual ports and poudriere. There's no guarantee that `TEST_DEPENDS` are built by the time _this port_ builds. At minimum we need `TEST_DEPENDS` added into the port's build dependency graph, but that won't work in all cases due to circular dependencies. We need a lot of special casing and workdir saving and a lot more complexity to let this work.

#355 explains most of the problems around this.

PR #1360, opened 2026-06-17
Implement "tailed log" functionality.

tjemg:
When clicking on the package status, do not download the log file, but open a new window that dynamically shows the current log in the browser window, while still allowing to download the entire log file.


bdrewery:
Can you include a screenshot?

tjemg:
Sure :-)
Here it is
tailed_log

bdrewery:
Very cool. I don't have time to review this today but will get it in once I do.

PR #1361, opened 2026-06-18
Improve the effectiveness of the fetching binary packages

alexdupre:
There are two issues in the current implementation:
1. sometimes poudriere fetches a package and then immediately (and wrongly) deletes it, causing the unneeded rebuild of many packages
2. sometimes poudriere deletes a stale package and doesn't fetch it again, causing again unneeded rebuilds

They are clearly not expected behaviors, because by stopping and re-starting poudriere the behavior eventually converge to the correct one. The two commits fix the above issues.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -5008,6 +5008,36 @@ download_from_repo_post_delete() {
 	bset "stats_fetched" "${_read_file_lines_read:?}"
 }
 
+# download_from_repo() rewrites pkg_fetch on each call. prepare_ports() may
+# fetch more than once (an initial pass plus a re-fetch of packages the
+# incremental checks deleted), so the fetched lists are accumulated into
+# pkg_fetch_all and reduced back into pkg_fetch by download_from_repo_finalize().
+download_from_repo_accumulate() {
+	[ $# -eq 0 ] || eargs download_from_repo_accumulate
+	if [ -s "${MASTER_DATADIR:?}/pkg_fetch" ]; then
+		cat "${MASTER_DATADIR:?}/pkg_fetch" \
+		    >> "${MASTER_DATADIR:?}/pkg_fetch_all"
+	fi
+}
+
+# Reduce pkg_fetch to the deduplicated set of fetched packages that still
+# exist (a package fetched in an earlier pass may have been deleted by the
+# incremental checks), for download_from_repo_post_delete().
+download_from_repo_finalize() {
+	[ $# -eq 0 ] || eargs download_from_repo_finalize
+	local pkgname
+
+	if [ -s "${MASTER_DATADIR:?}/pkg_fetch_all" ]; then
+		sort -u "${MASTER_DATADIR:?}/pkg_fetch_all" |
+		    while mapfile_read_loop_redir pkgname; do
+			if [ -e "${PACKAGES:?}/All/${pkgname}.${PKG_EXT}" ]; then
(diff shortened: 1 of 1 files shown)


src/share/poudriere/common.sh
```diff
@@ -10394,20 +10465,22 @@ prepare_ports() {
...
-				if sanity_check_pkgs; then
-					break
-				fi
-			done
-		else
-			msg "Skipping recursive rebuild"
-		fi
+		# delete_old_pkgs()/sanity_check_pkgs() may delete stale packages
+		# that were already present at the first download_from_repo().
+		# Re-fetch those instead of rebuilding them by doing a second pass.
```
bdrewery, line 10471: This violates the incremental build algorithm. It doesn't matter that we got a package from _fetching_ or _previous build_. If `delete_old_pkgs` deletes something it should stay deleted.

`download_from_repo_post_delete` has some comments and handling about this. https://github.com/freebsd/poudriere/blob/master/src/share/poudriere/common.sh#L4969-L4990

In the default case it can't do anything about it because of the incremental delete algorithm being too aggressive. In the `PKG_NO_VERSION_FOR_DEPS` case it asserts that nothing was fetched and then deleted because it can make smarter fetch decisions in that case.

bdrewery (changes requested):

bdrewery:
This is no different than making `bulk -S` the default (which is broken by design). It doesn't matter if a package is pre-existing or fetched or sneaked back in from a previous build after `delete_old_pkgs` deletes it. If `delete_old_pkgs` deletes it then we need to keep it deleted.

The real problem here isn't deleting _fetched_ packages, it is _deleting packages_. It's the incremental algorithm.

https://github.com/freebsd/poudriere/blob/master/docs/pkg_no_version_for_deps.md and https://github.com/freebsd/poudriere/blob/master/docs/shlib_tracking.md talk about the build algorithm problems here and attempts to solve it.

If we solve _deleting packages_ we don't need all of this fetch rehandling and can assert nothing is fetched+deleted.

Thank you for attempting to improve the situation. I don't see any way this can be committed without causing people to have broken packages so am closing it.

src/share/poudriere/common.sh
alexdupre: Hmmm, the idea of the patch was to fetch a *newer* package (not the same one), that was not fetched before because of the presence of a *stale* version. But I could be wrong.

alexdupre:
My patches were probably not good (first time I looked at poudriere source code), but surely something is wrong if restarting poudriere multiple times changes its behavior. Do we agree on this? What's your recommended way to use poudriere with binary packages?

bdrewery:
> My patches were probably not good (first time I looked at poudriere source code), 

Your patches were actually very good, it's just the change itself is problematic. I hate to discourage contributions by closing a PR like this.


> but surely something is wrong if restarting poudriere multiple times changes its behavior. Do we agree on this? What's your recommended way to use poudriere with binary packages?

I don't have a good answer right now. I agree the current situation isn't ideal at all. `delete_old_pkgs` is far too aggressive.

I had been pushing for this `PKG_NO_VERSION_FOR_DEPS` thing that avoids the aggressive rebuild and avoids fetch-deletes. But it does not work well with multi-repo setups for `pkg` and has some flaws around binary file dependencies.

The root problem is still the aggressive deleting. Nobody likes that and it is burning a lot of resources.

PR #1363, opened 2026-08-11
Reject unsafe PKGNAME values before package staging

acts-1631:
Poudriere uses PKGNAME in host-side package staging paths. A port
Makefile can supply path separators or traversal components, causing
those operations to escape the staging directory.

Validate PKGNAME after metadata extraction and again before a build
worker uses it. This prevents unsafe port metadata from reaching
host filesystem operations.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -5445,6 +5445,9 @@ build_port() {
 	local JEXEC_SETSID
 	local -
 
+	pkgname_is_valid "${pkgname}" ||
+	    err 1 "build_port: Invalid PKGNAME '${pkgname}'"
+
 	_my_path mnt
 	_log_path log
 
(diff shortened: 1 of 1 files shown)


src/share/poudriere/common.sh
```diff
@@ -6688,6 +6691,16 @@ _build_pkg_fp() {
 	fi
 }
 
+pkgname_is_valid() {
+	[ $# -eq 1 ] || eargs pkgname_is_valid pkgname
+	local pkgname="$1"
+
+	case "${pkgname}" in
+	""|*[!A-Za-z0-9+_,.~:-]*) return 1 ;;
```
bdrewery: Rather than an allowlist to maintain (in 2 places) can't we just denylist `/`?

bdrewery (changes requested):

src/share/poudriere/common.sh
acts-1631: Updated to a denylist. In addition to `/`, it rejects empty values and `.`/`..`, since `..` is a traversal component without a path separator.

Issue #1344, opened 2026-05-09
bulk idle builders leave their tmpfs mounted which wastes space


bdrewery:
From Mark Millard:
```
# poudriere version
poudriere-git-3.4.99.20260425

The following is an extraction of tmpfs lines from a df -m ouptput in a
UFS system context. The only active builder was 03 . All other future
builders were waiting for 03 to finish because they were dependent on
03's results.

1M-
blocks   Used  Avail Capacity  Mounted on
131780   2329 129451     2%
/usr/local/poudriere/data/.m/main-CA76-default/03
131442   1991 129451     2%
/usr/local/poudriere/data/.m/main-CA76-default/03/usr/local
131334   1883 129451     1%
/usr/local/poudriere/data/.m/main-CA76-default/06/usr/local
131144   1693 129451     1%
/usr/local/poudriere/data/.m/main-CA76-default/01/usr/local
130975   1524 129451     1%
/usr/local/poudriere/data/.m/main-CA76-default/06
130681   1230 129451     1%
/usr/local/poudriere/data/.m/main-CA76-default/01
130672   1221 129451     1%
/usr/local/poudriere/data/.m/main-CA76-default/02
130669   1218 129451     1%
/usr/local/poudriere/data/.m/main-CA76-default/04
130667   1216 129451     1%
/usr/local/poudriere/data/.m/main-CA76-default/08
130667   1216 129451     1%
/usr/local/poudriere/data/.m/main-CA76-default/07
130666   1215 129451     1%
/usr/local/poudriere/data/.m/main-CA76-default/ref
130666   1215 129451     1%
/usr/local/poudriere/data/.m/main-CA76-default/05
129848    397 129451     0%
/usr/local/poudriere/data/.m/main-CA76-default/02/usr/local
129845    393 129451     0%
/usr/local/poudriere/data/.m/main-CA76-default/04/usr/local
129557    106 129451     0%
/usr/local/poudriere/data/.m/main-CA76-default/07/usr/local
129548     97 129451     0%
/usr/local/poudriere/data/.m/main-CA76-default/08/usr/local
129451      0 129451     0%
/usr/local/poudriere/data/.m/main-CA76-default/ref/var/db/ports
129451      0 129451     0%
/usr/local/poudriere/data/.m/main-CA76-default
  2048     54   1993     3%
/usr/local/poudriere/data/.m/main-CA76-default/ref/.p
  2048     47   2000     2%
[...]

PR #1364, opened 2026-08-11
Confine distfile cache imports to their roots

acts-1631:
Poudriere imports distfiles from the build jail into DISTFILES_CACHE
as root. Port-controlled distfile paths and symlinks can resolve
outside their intended roots during that import.

Reject unsafe relative paths and verify resolved source and destination
directories remain below their intended roots before importing files.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -5267,6 +5267,18 @@ check_fs_violation() {
 	return $ret
 }
 
+distfile_path_is_safe() {
+	[ $# -eq 1 ] || eargs distfile_path_is_safe path
+	local path="$1"
+
+	case "${path}" in
+	""|/*|*/|*//*|.|..|./*|../*|*/./*|*/../*|*/.|*/..)
+		return 1
+		;;
+	esac
+	return 0
+}
+
 gather_distfiles() {
 	[ $# -eq 7 ] || [ $# -eq 6 ] ||
 	    eargs gather_distfiles '[-l]' originspec_main pkgname_main \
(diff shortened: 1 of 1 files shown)


src/share/poudriere/common.sh
```diff
@@ -5308,11 +5321,30 @@ gather_distfiles() {
...
+	esac
+	for d in ${dists}; do
+		if ! distfile_path_is_safe "${d}"; then
+			err 1 "gather_distfiles: Invalid distfile path '${d}'"
+		fi
+	done
 
 	job_msg_dev "${COLOR_PORT}${origin}${flavor:+@${flavor}}${subpkg:+~${subpkg}} | ${pkgname_main}${COLOR_RESET}: distfiles ${from} -> ${to}"
 	mkdir -p "${to}/${sub}"
+	dstroot="$(realpath -q "${to}/${sub}")" ||
```
bdrewery, line 5340: I wonder if we need all of the extra `realpath` handling since we already check for bad patterns above.

acts-1631: The pattern checks are deliberately only lexical. The escape remains possible through an existing intermediate symlink: a benign-looking `DIST_SUBDIR` or nested `ALLFILES` parent can resolve outside its configured root. `dstroot` protects `DIST_SUBDIR`, `srcpath_real` prevents importing an out-of-tree source link, and `dstpath_parent_real` catches a nested destination symlink. Removing those checks would restore the symlink escape.

acts-1631:
Friendly ping for this one and the others when you have time. @bdrewery

bdrewery:
I've sat on this because I'm hesitant to place symlink restrictions on the distfiles dir as the user may have created some, same as the package repo concern I may not have said. And symlinks outside of the distfile dir are intentionally used. So I need to think about it and see how it interacts with some other pending changes for #1344.

PR #1365, opened 2026-08-11
Reject client-supplied poudriered log paths

acts-1631:
Poudriered accepts commands over its local Unix socket, authorizes the `command` and `arguments` fields, then retains the complete UCL request in its queue. `execute_cmd()` later used an unvalidated client `log` field to create parent directories and open the path as root with `O_CREAT|O_TRUNC`.

A low-privilege user authorized for an otherwise harmless command could therefore create or truncate root-owned files at an attacker-selected path. The `/tmp/poudriered.log` fallback also allowed a pre-created symlink to redirect the root open.

Reject log paths supplied by clients. Keep explicit paths available for trusted scheduled commands, use `/var/run/poudriered.log` as the fallback, and use `O_NOFOLLOW` when opening logs.

### DIFF
--- a/src/poudriered/poudriered.c
+++ b/src/poudriered/poudriered.c
@@ -491,21 +491,23 @@ execute_cmd(void)
 	int error;
 	char **argv;
 	char *buf, *tofree, *arg;
+	const char *logpath;
 	int argc, argvl;
 	const ucl_object_t *o, *a, *l;
 
 	if (running == NULL)
 		return;
 
+	logpath = "/var/log/poudriered.log";
 	l = ucl_object_find_key(running, "log");
-	if (l != NULL)
-		mkdirs(ucl_object_tostring(l), true);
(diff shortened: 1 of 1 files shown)


src/poudriered/poudriered.c
```diff
@@ -491,21 +491,23 @@ execute_cmd(void)
...
 	char **argv;
 	char *buf, *tofree, *arg;
+	const char *logpath;
 	int argc, argvl;
 	const ucl_object_t *o, *a, *l;
 
 	if (running == NULL)
 		return;
 
+	logpath = "/var/run/poudriered.log";
```
bdrewery: /var/log is typically where logs go rather than /var/run.

acts-1631: Updated.

acts-1631:
@bapt Hi, would you consider looking at my open pull requests on this repo too?

PR #1366, opened 2026-08-11
Constrain package-output symlinks before host import

acts-1631:
The package phase exposes a host-backed `/.npkg` directory to the jail build user. `build_port()` previously moved every file and symlink from that directory into the host package repository without checking the output layout or a symlink target.

An absolute symlink created by code running during a build has jail-relative meaning while `/.npkg` is mounted, but resolves on the host after import. It persists through the default atomic repository handoff. On the next `bulk` run, default unknown-file cleanup calls `realpath()` on the repository entry and `delete_pkg()` unlinks that resolved target as root.

A compromised upstream distfile or third-party build source can therefore cause arbitrary host-file deletion. No malicious ports metadata or ports-tree commit is required.

Only import direct `All/` and `Latest/` output entries, and reject symlinks whose resolved target is outside the per-build staging directory. Also restrict hashed-package cleanup to targets below the repository-owned `All/Hashed` directory, preserving legitimate `pkg repo --hash --symlink` cleanup without following an unexpected link.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -5443,6 +5443,7 @@ build_port() {
 	local max_execution_time allownetworking
 	local NEED_ROOT PREFIX MAX_FILES
 	local JEXEC_SETSID
+	local pkg_stage pkg_stage_real pkg_target
 	local -
 
 	pkgname_is_valid "${pkgname}" ||
@@ -5882,13 +5883,40 @@ build_port() {
 		esac
 	done
 
-	if [ -d "${PACKAGES}/.npkg/${pkgname}" ]; then
+	pkg_stage="${PACKAGES}/.npkg/${pkgname}"
+	if [ -d "${pkg_stage}" ]; then
 		# everything was fine we can copy the package to the package
 		# directory
-		find "${PACKAGES:?}/.npkg/${pkgname}" \
-		    -mindepth 1 \( -type f -or -type l \) |
+		# Only the pkg bootstrap needs package-manager compatibility links.
+		pkg_stage_real=
+		case "${port}" in
+		ports-mgmt/pkg|ports-mgmt/pkg-devel)
+			pkg_stage_real="$(realpath -q "${pkg_stage}")" ||
+			    err 1 "build_port: realpath ${pkg_stage}"
+			;;
+		esac
+		{
+			case "${pkg_stage_real}" in
+			"") ;;
+			*)
+				# Process links first so their targets are still in staging.
+				find "${pkg_stage}" -mindepth 1 -type l
+				;;
+			esac
+			find "${pkg_stage}" -mindepth 1 -type f
+		} |
 		    while mapfile_read_loop_redir pkg_path; do
-			pkg_file="${pkg_path#"${PACKAGES}/.npkg/${pkgname:?}"}"
+			if [ -L "${pkg_path}" ]; then
+				pkg_target="$(realpath -q "${pkg_path}" || :)"
+				case "${pkg_target}" in
+				"${pkg_stage_real:?}/"*) ;;
+				*)
+					msg_warn "Ignoring package symlink outside staging: ${pkg_path}"
+					continue
+					;;
+				esac
+			fi
+			pkg_file="${pkg_path#"${pkg_stage}"}"
 			pkg_base="${pkg_file%/*}"
 			mkdir -p "${PACKAGES:?}/${pkg_base:?}"
 			# rename as this is expected to be on the same


src/share/poudriere/include/pkg.sh
```diff
@@ -490,12 +490,27 @@ delete_all_pkgs() {
...
 	local pkg="$1"
+	local pkg_all_dir pkg_hashed_dir pkg_target
 
 	clear_pkg_cache "${pkg}"
 
-	# If ${pkg} is a symlink, delete the target as well
+	# Hashed repositories use symlinks from All/ to All/Hashed/.  Do not
+	# unlink a target outside that repository-owned directory.
 	if [ -L "${pkg}" ]; then
-		unlink "$(realpath "${pkg}")"
```
bdrewery, line 498: This code isn't intended only for Hashed mode. It also covers older `package.txz` -> `package.pkg`.
But in general I am not sure this is worth the effort. If someone can place a malicious symlink in the package repo and lead to a file being deleted, isn't the system/repository already compromised?

acts-1631: I see `delete_pkg()` must also handle legacy `package.txz -> package.pkg` links, so constraining it to `All/Hashed` is too narrow indeed. I'll remove that part.

The symlink does not require host write access to the package repository. During the package phase, `build_port()` mounts `${PACKAGES}/.npkg/${pkgname}` at `/.npkg`, chowns it to the jail build user, then imports links written there into `${PACKAGES}`. Code running during a build can therefore introduce an absolute link that is interpreted against the jail root while mounted, but against the host root after import.

If code running in a jailed port build is considered trusted for this host-output boundary, I can close the PR. Otherwise I will reduce the fix to validating imported symlink targets against their per-build staging root and preserving normal repository link formats. Let me know what you think.

src/share/poudriere/common.sh
```diff
@@ -5879,15 +5880,50 @@ build_port() {
...
-	if [ -d "${PACKAGES}/.npkg/${pkgname}" ]; then
+	pkg_stage="${PACKAGES}/.npkg/${pkgname}"
+	if [ -d "${pkg_stage}" ]; then
+		pkg_stage_real="$(realpath -q "${pkg_stage}")" ||
+		    err 1 "build_port: realpath ${pkg_stage}"
 		# everything was fine we can copy the package to the package
 		# directory
-		find "${PACKAGES:?}/.npkg/${pkgname}" \
+		find "${pkg_stage}" \
 		    -mindepth 1 \( -type f -or -type l \) |
```
bdrewery: I think we don't need to copy symlinks outside of ports-mgmt/pkg, which can simplify a lot here.

acts-1631: I simplified it a bit per your suggestion. How does this one look?

bdrewery:
Thanks, this is easier to work with. I didn't like enforcing rules on existing repositories.

We may be able to reduce the window that .npkg is mounted and writable as well to be limited to the `make package` phase when `$INTERACTIVE_MODE == 0`. We would need to revert the `PACKAGES=/.npkg` line after taking the file into the real package dir. That doesn't have to happen here.

I am running tests on the change now.

src/share/poudriere/common.sh
```diff
@@ -5885,8 +5885,17 @@ build_port() {
 	if [ -d "${PACKAGES}/.npkg/${pkgname}" ]; then
 		# everything was fine we can copy the package to the package
 		# directory
-		find "${PACKAGES:?}/.npkg/${pkgname}" \
-		    -mindepth 1 \( -type f -or -type l \) |
+		# Only the pkg bootstrap needs package-manager compatibility links.
+		case "${port}" in
+		ports-mgmt/pkg)
```
bdrewery: I'm having trouble with the tests but we probably need `ports-mgmt/pkg-devel` here as well.

bdrewery: The extra logic to constrain the symlink target in this branch makes sense.

acts-1631: Good catch on `pkg-devel`. I have added it alongside `pkg`, and restored the target check for the links those ports are allowed to emit. The links are handled before their target files so the normal bootstrap link is still valid when checked. Everything else still imports regular files only.

src/share/poudriere/common.sh
```diff
@@ -5882,13 +5883,42 @@ build_port() {
...
-		    -mindepth 1 \( -type f -or -type l \) |
+		# Only the pkg bootstrap needs package-manager compatibility links.
+		case "${port}" in
+		ports-mgmt/pkg|ports-mgmt/pkg-devel)
+			pkg_stage_real="$(realpath -q "${pkg_stage}")" ||
+			    err 1 "build_port: realpath ${pkg_stage}"
+			;;
+		esac
+		case "${port}" in
+		ports-mgmt/pkg|ports-mgmt/pkg-devel)
```
bdrewery: Is there a reason these are split?

src/share/poudriere/common.sh
```diff
@@ -5882,13 +5883,42 @@ build_port() {
...
+			# Process links first so their targets are still in staging.
+			{
+				find "${pkg_stage}" -mindepth 1 -type l
+				find "${pkg_stage}" -mindepth 1 -type f
+			}
```
bdrewery: I don't understand the comment or why we're doing 2 finds now. Elaborate?

acts-1631: The two parts are related to the target check. For the bootstrap ports, `find` normally visits `All` before `Latest`, so the regular package file is moved first and a valid `Latest/pkg.txz -> ../All/pkg.txz` link becomes dangling in staging before it can be checked. The first `find` emits links so their targets can be validated while still in staging; the second imports regular files.

The staging-root assignment is separate because the `while` runs on the right side of a pipeline, so it has to be set before the pipeline. I can refactor the duplicated port match if you would prefer.

bdrewery: Got it, that all makes sense now. The `{ }` inside the case statement shouldn't be needed. Moving them out could make the `pkg_stage_real` assignment more clearly outside of the pipe.

acts-1631: Okay

bdrewery (approved):
Code looks good. Running tests on it now.

bdrewery:
Thanks for your patience. I'm copying these over to the release branch as they are merged.

PR #1370, opened 2026-08-27
Restrict package stage output names

acts-1631:
build_port() mounts the package output directory at /.npkg in the jail,
then imports every regular file from that directory into the shared
package repository.

A process started by code during configure, build, or test can remain
alive until the package phase. Once /.npkg is mounted, it can write an
archive named after an unrelated package. The host-side importer then
renames that archive into the repository, replacing an existing package
with the same name before the jail is cleaned up.

This does not make a compromised port's own package trustworthy. The
problem is that code from one build can silently alter other package
archives in the shared repository.

Only promote All/<pkgname>.<suffix> files declared by PKGNAMES for the
origin. This retains legitimate subpackage outputs while rejecting
unrelated package archives.

### DIFF
--- a/src/share/poudriere/common.sh
+++ b/src/share/poudriere/common.sh
@@ -5443,6 +5443,7 @@ build_port() {
 	local max_execution_time allownetworking
 	local NEED_ROOT PREFIX MAX_FILES
 	local JEXEC_SETSID
+	local expected_pkgnames expected_pkgname pkg_output_name
 	local pkg_stage pkg_stage_real pkg_target
 	local -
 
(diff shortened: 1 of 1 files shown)


bdrewery:
I still intend to get back to these and appreciate your PRs. Finding time here is difficult.

acts-1631:
> I still intend to get back to these and appreciate your PRs. Finding time here is difficult.

Thank you for your effort. I decided to close the PRs due to a recent dispute with the FreeBSD core team in which they decided to take several security fixes I sent with attribution intentionally removed. I did not have that problem here or in the pkg repo, but I decided to stop contributing to anything FreeBSD as a result. I appreciate the time you spent reviewing the merged fixes.

bdrewery:
That's unfortunate. Could it be an oversight? They are overwhelmed too.

acts-1631:
> That's unfortunate. Could it be an oversight? They are overwhelmed too.

It was not an oversight, but a long conversation via email. The maintainer of fetch and libfetch refused the attribution line in my patches (2 months after they were sent in) because my username is a reference to a Bible verse. I asked to either attribute properly or simply drop the patches and everyone move on. He refused, so I asked core to intervene. After some back-and-forth, they sided with him and I was told they would take the fixes without mention of the author or reporter. That was my first and last security submission to FreeBSD.

PR #1378, opened 2026-09-04
image: Use quiet-aware msg idiom for pkgbase install status

michael-o:
install_world_from_pkgbase() always reported "Installing base packages" / "Base packages installed" as two separate msg() lines, regardless of PKG_QUIET. Under the default quiet pkg install (-q, unless -v was passed), nothing prints between them, so treat it like the other silent/slow operations with msg_n "...done". When -v clears PKG_QUIET, pkg's own install output streams between the header and completion, so keep the header on its own line via msg() with a trailing "..." and skip the redundant "done", matching the quiet-vs-verbose idiom already used for jail.sh's install_from_vcs().


src/share/poudriere/image.sh
```diff
@@ -250,12 +250,16 @@ install_world_from_pkgbase()
...
+		msg_n "Installing base packages..."
+	else
+		msg "Installing base packages..."
+	fi
 	while read line; do
 		pkg -o ABI_FILE="${mnt}/usr/lib/crt1.o" -o REPOS_DIR=${WRKDIR}/world/etc/pkg/ -o ASSUME_ALWAYS_YES=yes -r ${WRKDIR:?}/world install -r local ${PKG_QUIET} -y ${line}
 	done < ${PKGBASELIST}
 	rm ${WRKDIR:?}/world/etc/pkg/FreeBSD-base.conf
-	msg "Base packages installed"
+	[ -n "${PKG_QUIET}" ] && echo " done"
```
bdrewery: This creates a trap for the return status. If `PKG_QUIET` is set then the function returns a failure. If `||` or `if ... then... fi` to avoid.

bdrewery (changes requested):

src/share/poudriere/image.sh
michael-o: Noted, to be fair this needs to be fixed as well: https://github.com/freebsd/poudriere/commit/4cb789623252ddaee50a0acd90f02d0f3eea3943

michael-o: Done.

michael-o: > Noted, to be fair this needs to be fixed as well: [4cb7896](https://github.com/freebsd/poudriere/commit/4cb789623252ddaee50a0acd90f02d0f3eea3943)

Picked up followup fix in: https://github.com/freebsd/poudriere/pull/1387

bdrewery (approved):

michael-o:
Merged.

PR #1382, opened 2026-09-06
Add missing FLAVOR to stage-qa and check-plist calls

svmhdvn:
For example, `poudriere testport -NNi devel/freebsd-gcc15@amd64` will correctly build the amd64 flavor, but will attempt to stage-qa and check-plist for the aarch64 flavor because it becomes the default flavor through alphabetic sorting.


svmhdvn:
Might be related to https://github.com/freebsd/poudriere/pull/1371 and could have many other missing `FLAVOR` vars in other places.

michael-o:
> Might be related to #1371 and could have many other missing `FLAVOR` vars in other places.

I have scanned the rest. Didn't find anything.

PR #1383, opened 2026-09-09
Make bulk resilient to single-port errors during prepare

mat813:
A single port with bad metadata used to abort the entire bulk run:

- an invalid FLAVOR (requested directly or via a dependency),
- a dependency on a nonexistent origin, a self-dependency,
- a malformed origin spec,
- an empty FLAVOR spec,
- an unresolvable dependency,
- a dead package in the queue graph.

Now just that port is ignored, with a warning, like a blacklisted port.

`-D` restores the old fatal behavior. testport keeps it by default, since it only tests one specific port.


bdrewery:
Do tests pass? `make -j2 check` in poudriere git.

Also what's the justification for default and always as opposed to limited to `bulk -a` or opt-in? This would be a surprising change.

mat813:
> Do tests pass? `make -j2 check` in poudriere git.

I did not think to check if there were test.
They don't start right now as it tries to use 13.5, and it is not available on downloads.f.o any more, I'm patching prep.sh to use 14.4, and launching `make -C test prep`, I'll look at what's what after dinner.

> Also what's the justification for default and always as opposed to limited to `bulk -a` or opt-in? This would be a surprising change.

The idea is that poudriere is here to build packages, and bulk should, by default, try to build packages, and not fail for things that are not fatal.
In the past few months, we've had weeks without package builds because a port was broken in a way or the other, and when it was fixed, another had some spec of dust that was not absolutely right, and so on.
A few ports being broken should not prevent the build of packages.
A developer who is working on things will either see packages missing after the run, or notice the warnings, and can pass -`D`. Even for a developer, I think having things that can be built start to build is a time gain as deps start building while you can work on your broken ports.

mat813:
I fixed the tests that were not already failing for me on master.

bdrewery:
> The idea is that poudriere is here to build packages, and bulk should, by default, try to build packages, and not fail for things that are not fatal.

But this is in conflict with another perspective. If I ask poudriere to build a set of packages I expect it to build those packages and not silently skip them. "silent" isn't the right word. It's more in an automated run of a script without errors, against packages in a file, it would continue shipping missing packages until someone bothered to read the log.

> In the past few months, we've had weeks without package builds because a port was broken in a way or the other, and when it was fixed, another had some spec of dust that was not absolutely right, and so on.

Perhaps these changes you are proposing make sense as a default for `bulk -a`, and we leave non-a alone?

mat813:
Changed things to make it non fatal for -a/-f, and fatal in all other cases.

PR #1386, opened 2026-09-13
ports, jail: Warn instead of erroring on hyphenated PTNAME/JAILNAME

michael-o:
Downgrade the ports-tree hyphen check from a hard err() to a warning,
consistent with the existing SETNAME hyphen warning. Extract the
hyphen check into a shared check_name_hyphen() helper in common.sh and
apply it uniformly at the point each identifier is established:
jail creation, jail rename, and ports tree creation. SETNAME keeps its
existing call in jail_start(), since a set has no separate creation
step -- its first use in a build is effectively its creation.

The warning intentionally doesn't name a specific mechanism (e.g. the
previous "-z" claim for ports tree names): no code path splits a
hyphenated name back into components during normal creation or build
handling. The only place a joined name is ever decomposed is
for_each_build()'s fallback for build logs predating separate
jailname/ptname/setname metadata (2013, a2300607), which no longer
applies to any name established from this point forward. Warning only
at creation time reflects that once a name exists, cautioning about it
again on every later build is not actionable.


src/share/poudriere/common.sh
```diff
@@ -3859,7 +3859,11 @@ jail_start() {
 
 	case "${setname}" in
 	*-*)
-		msg_warn "Using '-' in a SETNAME is not recommended as it causes ambiguities with parsing the build name of ${MASTERNAME}"
+		# XXX: This is only ambiguous for build logs from before
+		# jailname/ptname/setname were recorded as separate metadata
+		# (2013). Remove this warning around 2027-03 once any such
+		# logs are long gone from normal log rotation.
```
bdrewery: Hard disagree on this. How do these disambiguate?

```
   Create optional poudriere.conf
     You can also specify an optional poudriere.conf that is pulled in depending on the build.  Any of the following are allowed and will all be used in
     the order shown:

           /usr/local/etc/poudriere.conf
           /usr/local/etc/poudriere.d/poudriere.conf
           /usr/local/etc/poudriere.d/<setname>-poudriere.conf
           /usr/local/etc/poudriere.d/<tree>-poudriere.conf
           /usr/local/etc/poudriere.d/<jailname>-poudriere.conf
           /usr/local/etc/poudriere.d/<tree>-<setname>-poudriere.conf
           /usr/local/etc/poudriere.d/<jailname>-<tree>-poudriere.conf
           /usr/local/etc/poudriere.d/<jailname>-<setname>-poudriere.conf
           /usr/local/etc/poudriere.d/<jailname>-<tree>-<setname>-poudriere.conf

     Running poudriere with flag -v will show which files are included.

   Create optional make.conf
     You can also specify a global make.conf which will be used for all the jails.  Any of the following are allowed and will all be used in the order
     shown:

           /usr/local/etc/poudriere.d/make.conf
           /usr/local/etc/poudriere.d/<setname>-make.conf
           /usr/local/etc/poudriere.d/<tree>-make.conf
           /usr/local/etc/poudriere.d/<jailname>-make.conf
           /usr/local/etc/poudriere.d/<tree>-<setname>-make.conf
           /usr/local/etc/poudriere.d/<jailname>-<tree>-make.conf
           /usr/local/etc/poudriere.d/<jailname>-<setname>-make.conf
           /usr/local/etc/poudriere.d/<jailname>-<tree>-<setname>-make.conf
           /usr/local/etc/poudriere.d/hooks/plugins/<plugin>/make.conf

```

michael-o: Before I look at your objection. Is your objection regarding the demotion or the comment itself?

michael-o:
Thanks for the pushback — I looked closer at the mechanism you're pointing at (setup_makeconf/poudriere.conf's <jailname>-<tree>-<setname> file naming) and you're right that this isn't purely a
  legacy-log-parsing concern. But I also found that JAILNAME has never had a hyphen restriction at all, only PTNAME did — so if that config-file collision is the real risk, blocking hyphens in PTNAME alone was
  already inconsistent and non-load-bearing; JAILNAME was always the wide-open side of the same concatenation.

  Rather than resolve that by re-arguing scope, I reworked the change to close the actual gap:

  - Extracted the hyphen check into a shared check_name_hyphen() helper in common.sh.
  - Applied it consistently at the point each identifier is established: jail creation, jail rename (NEWJAILNAME had no validation at all before this), and ports tree creation. SETNAME keeps its existing check
  in jail_start(), since a set's first use in a build is effectively its creation — there's no separate -c step for sets.
  - Dropped the "only ambiguous for pre-2013 build logs, remove ~2027-03" framing and the "-z parsing" claim from the message text. Neither holds up as a universal explanation: the only place a joined name is
  ever decomposed back into components is for_each_build()'s fallback for build logs that predate separate jailname/ptname/setname metadata (2013, a2300607) — that's real but narrow, and doesn't justify
  asserting a specific parsing mechanism in the warning text itself.
  - Left the dot/period validation untouched — it has its own gaps (e.g. missing on jail rename) but that's a separate issue from this PR's scope.

  Net effect: JAILNAME, PTNAME, and SETNAME all get the same warning, at the same "creation" semantics, with no overclaiming about why. Warning only fires at creation time (not on every subsequent build) since
  once a name exists, repeating the caution on every build isn't actionable.

  Let me know if this addresses the concern or if there's still a gap I'm missing.

michael-o:
@bdrewery Did you find any time to review my changes?

PR #1390, opened 2026-09-17
tests: use 14.5-RELEASE for test jail

svmhdvn:
This fixes `make check`.


brd:
Which tests are you seeing change?  When I run make check it does not change the output:

poudriere w/ out patch
```
# TOTAL: 230
# PASS:  94
# SKIP:  133
# XFAIL: 1
# FAIL:  2
# XPASS: 0
# ERROR: 0
```

w/ patch
```
# TOTAL: 230
# PASS:  94
# SKIP:  133
# XFAIL: 1
# FAIL:  2
# XPASS: 0
# ERROR: 0
```

I am guessing the skipped ones?

svmhdvn:
Yes, try running as root.

test/common.bulk.sh
```diff
@@ -1522,7 +1522,7 @@ ARCH=$(uname -p)
 # and makes debugging multiple worktrees difficult. Just hash the srcdir
 # into the name.
 JAILNAME="poudriere-test-${ARCH}-$(realpath "${am_abs_top_srcdir:?}" | sha256 | cut -c1-6)"
-JAIL_VERSION="13.5-RELEASE"
+JAIL_VERSION="15.1-RELEASE"
```
michael-o: Why not the oldest supported version?

svmhdvn: I guess that's reasonable as well, up to you. I always like testing with the latest versions to proactively catch issues that might show up in newer releases. I don't mind either way though.

svmhdvn: Went with 14.5 since 14.4 is expected to be EOL in a few months.

Issue #1236, opened 2025-07-22
poudriere fails to build from ports


void-fm:
## Prerequisites
- [ ] Have you checked for an existing issue describing your problem?
- [ ] yes
- [ ] Is your ports tree recent?
- [ ] yes

## Describe the bug
poudriere fails to build

fails here:

cd external/sh &&  /usr/ports/ports-mgmt/poudriere/work/poudriere-3.4.2/external/sh/mknodes  /usr/ports/ports-mgmt/poudriere/work/poudriere-3.4.2/external/sh/nodetypes  /usr/ports/ports-mgmt/poudriere/work/poudriere-3.4.2/external/sh/nodes.c.pat
cd external/sh &&  /usr/ports/ports-mgmt/poudriere/work/poudriere-3.4.2/external/sh/mksyntax
cat external/sh/builtins.def src/poudriere-sh/builtins-poudriere.defexternal/sh/builtins.def src/poudriere-sh/builtins-poudriere.def > src/poudriere-sh/builtins.def
cat: src/poudriere-sh/builtins-poudriere.defexternal/sh/builtins.def: No such file or directory
*** Error code 1

Stop.
make[1]: stopped making "all-am" in /usr/ports/ports-mgmt/poudriere/work/poudriere-3.4.2
*** Error code 1

Stop.
make: stopped making "all" in /usr/ports/ports-mgmt/poudriere/work/poudriere-3.4.2
*** Error code 1

Stop.
make[1]: stopped making "/usr/ports/ports-mgmt/poudriere/work/.stage_done.poudriere._usr_local" in /usr/ports/ports-mgmt/poudriere
*** Error code 1

Stop.
make: stopped making "all" in /usr/ports/ports-mgmt/poudriere

## How to reproduce
1.  cd /usr/ports/ports-mgmt/poudriere && make MAKE_JOBS_UNSAFE=yes

## Expected behavior
poudriere gets built

## Environment
 - Host OS [e.g. 12.2 amd64]: 15.0-CURRENT #0 n278914-e0838f8a2e61 GENERIC-NODEBUG amd64 1500053 1500053 
 - Ports branch and revision [e.g. 2020Q3 r550754]: # git -C /usr/ports rev-list --count --first-parent HEAD
713433

ghost:
This is caused by the recent bmake update in main (a8c56be47166).

void-fm:
> This is caused by the recent bmake update in main (a8c56be47166).

yikes! How to unbreak?

[...]

PR #1392, opened 2026-09-18
test-ports: backport bmake .ALLSRC fix to poudriere-devel

svmhdvn:


michael-o:
Is there an issue for this? What is the objective here?

svmhdvn:
This proactively fixes the test suite with test jails containing a bmake >= 20250630. This brings in the fix from https://github.com/freebsd/poudriere/issues/1236.

